Issue #20, opened 2019-12-03
emit JSON


cburgard:
There is no JSON emitter as far as I can tell. Are there any plans to add one? It should be a minimal addition, as far as I can tell.

biojppm:
I have no immediate plans to add one. I have other urgent things to look at, like the installer and conformance to the YAML test suite.

If you are in a rush, I would of course accept a PR implementing that functionality. The natural place to put that would be the [emit.hpp](https://github.com/biojppm/rapidyaml/blob/master/src/c4/yml/emit.hpp) header file. You can also take guidance from the code there.

cburgard:
Thanks for the guidance. I had a look at the code and found that the only things that would really need to be changed would be the contents of `emit.def.hpp`, whereas the `emit.hpp` would mostly be copied.
I feel like the most elegant solution would be to template the contents of `emit.hpp` to some , write template specializations of the contents of `emit.hpp` for YML and JSON type output, and then instantiate the writers in `emit.hpp` to the corresponding types.
Also, I'm tempted to move the whole JSON stuff to a separate namespace, perhaps `c4/json` to avoid confusion.
What do you think about these plans? Would you prefer things to be done differently? If so, how?

biojppm:
Something like what you say could work, with some adjustments. Let's focus first on making the code work, then we can refactor to tidy it up (regarding namespaces/headers/function names, etc).

There are three separate entities in that file: the writers, the emitter class and the emit() freestanding functions. 

 * The writers need no adjustment
 * For now, I would duplicate and work on the emitter class, which does most of the work, to adjust for emitting valid json. Let's call this EmitterJSON. Then at the top of the file we would have EmitterJSONBuf, etc. Since JSON is so much simpler than YAML, this class will be much shorter; the parts of YAML that are incompatible with JSON (eg, anchors and references) should be verified for on the visit functions. For now, let's make this an error if any such entity exists when emitting JSON.
* Finally, each freestanding function at the end of the file would have a emit_json() counterpart, placed next to it.

This will result in `c4::yml::emit_json()` which is indeed awkward, but we'll address that later.

The most important refactor will be in retaining the common parts of the emitter classes. But first, let's have it working; duplication there is no big deal since this is a transient situation, and when we have both yml and json emitters it will be obvious to see how to arrange the classes; maybe some form of CRTP would be best (and absolutely no virtual calls).

cburgard:
Ok, I'll write down some preliminary version soon then.
[...]

PR #22, opened 2019-12-04
JSON emitter

cburgard:
Added a first preliminary version of the JSON emitter #20


src/c4/yml/emit.def.hpp
```diff
@@ -268,4 +268,172 @@ void Emitter<Writer>::_write_scalar(csubstr s)
...
+    return this->Writer::m_pos;
+}
+
+/** @todo this function is too complex. break it down into manageable pieces */
+template<class Writer>
+void EmitterJSON<Writer>::_do_visit(Tree const& t, size_t id)
+{
+  if(t.is_doc(id))
+    {
+      throw "we don't do that yet";
```
biojppm: no exceptions; call the error callback. The client chooses whether he throws an exception, calls abort, or something else.

src/c4/yml/emit.def.hpp
```diff
@@ -268,4 +268,172 @@ void Emitter<Writer>::_write_scalar(csubstr s)
...
+        _write("...\n");
+    }
+    return this->Writer::m_pos;
+}
+
+/** @todo this function is too complex. break it down into manageable pieces */
+template<class Writer>
+void EmitterJSON<Writer>::_do_visit(Tree const& t, size_t id)
+{
+  if(t.is_doc(id))
```
biojppm: watchout for indentation style

src/c4/yml/emit.hpp
```diff
@@ -241,6 +240,228 @@ CharOwningContainer emitrs(NodeRef const& n)
...
+//
+///** emit JSON to an STL-like ostream
+// * @overload */
+//template<class OStream>
+//inline OStream& operator<< (OStream& s, NodeRef const& n)
+//{
+//    EmitterOStream<OStream> em(s);
+//    em.emit(n);
+//    return s;
+//}
```
biojppm: see this comment: https://github.com/biojppm/rapidyaml/issues/20#issuecomment-561701426

src/c4/yml/emit.hpp
```diff
@@ -241,6 +240,228 @@ CharOwningContainer emitrs(NodeRef const& n)
...
+    substr result = em.emit(t, id, error_on_excess);
+    return result;
+}
+
+/** emit JSON to the given buffer. Return a substr trimmed to the emitted JSON.
+ * Raise an error if the space in the buffer is insufficient.
+ * @overload */
+inline substr emit_json(Tree const& t, substr buf, bool error_on_excess=true)
+{
+    return emit(t, t.root_id(), buf, error_on_excess);
```
biojppm: should be emit_json()

src/c4/yml/emit.hpp
```diff
@@ -241,6 +240,228 @@ CharOwningContainer emitrs(NodeRef const& n)
...
+}
+
+/** emit JSON to the given buffer. Return a substr trimmed to the emitted JSON.
+ * Raise an error if the space in the buffer is insufficient.
+ * @overload
+ */
+inline substr emit_json(NodeRef const& r, substr buf, bool error_on_excess=true)
+{
+    return emit(*r.tree(), r.id(), buf, error_on_excess);
```
biojppm: should be emit_json()

src/c4/yml/emit.hpp
```diff
@@ -241,6 +240,228 @@ CharOwningContainer emitrs(NodeRef const& n)
...
+    return emitrs_json(t, t.root_id(), cont);
+}
+
+/** emit+resize: JSON to the given std::string/std::vector-like container,
+ * resizing it as needed to fit the emitted JSON. */
+template<class CharOwningContainer>
+CharOwningContainer emitrs_json(Tree const& t)
+{
+    CharOwningContainer c;
+    emitrs(t, t.root_id(), &c);
```
biojppm: should be emitrs_json()

src/c4/yml/emit.hpp
```diff
@@ -241,6 +240,228 @@ CharOwningContainer emitrs(NodeRef const& n)
...
+    return c;
+}
+
+/** emit+resize: JSON to the given std::string/std::vector-like container,
+ * resizing it as needed to fit the emitted JSON. */
+template<class CharOwningContainer>
+substr emitrs_json(NodeRef const& n, CharOwningContainer * cont)
+{
+    return emitrs(*n.tree(), n.id(), cont);
```
biojppm: should be emitrs_json()

src/c4/yml/emit.hpp
```diff
@@ -241,6 +240,228 @@ CharOwningContainer emitrs(NodeRef const& n)
...
+}
+
+/** emit+resize: JSON to the given std::string/std::vector-like container,
+ * resizing it as needed to fit the emitted JSON. */
+template<class CharOwningContainer>
+CharOwningContainer emitrs_json(NodeRef const& n)
+{
+    CharOwningContainer c;
+    emitrs(*n.tree(), n.id(), &c);
```
biojppm: should be emitrs_json()

src/c4/yml/emit.def.hpp
cburgard: I hope I got the right one!

src/c4/yml/emit.def.hpp
cburgard: I refactored the indentations in the entire document. Does it meet the projects style now?

src/c4/yml/emit.hpp
cburgard: done (slightly differently to incorporate both Trees and NodeRefs)

src/c4/yml/emit.hpp
cburgard: done

src/c4/yml/emit.hpp
cburgard: done

src/c4/yml/emit.hpp
cburgard: done

src/c4/yml/emit.hpp
cburgard: done

src/c4/yml/emit.hpp
cburgard: done

src/c4/yml/emit.def.hpp
biojppm: actually, no... compare to the original function.

biojppm:
Don't forget to add a unit test proving that this works.

src/c4/yml/emit.def.hpp
cburgard: next attempt

cburgard:
> Don't forget to add a unit test proving that this works.

I'm happy to add one, but I don't seem to understand how your test suite works. It seems to be using preprocessor macros mostly, and none of the tests are known to ctest. Can you point me to some documentation explaining your particular unit test suite style so I know how to add something?

For development, I was just using the example you have in the README, so I've attached
 my code FWIW.
[test.cpp.txt](https://github.com/biojppm/rapidyaml/files/3926074/test.cpp.txt)

biojppm:
Yes, most of the YAML test cases you'll see are doing full and deeply nested assertions about the nature of each element in the hierarchy (eg, elm foo/bar/baz is a map, has 3 value entries with contents "a", "b", "c", is tagged, has an anchor, etc.  So I implemented a way to do it declaratively; otherwise it would be too much work.

However, there are some simple tests you can pick up from in [the basic.cpp file](https://github.com/biojppm/rapidyaml/blob/master/test/basic.cpp).

I suggest adding a basic_json.cpp file. The first tests would create a ryml tree from parsing some JSON-compatible YAML, then emitting, then parsing the emitted result, emitting again, and verifying that the emitted strings are equal. This would check for sanity and at the very least YAML conformance.

Next you could refine by adding tests looking inside the emitted JSON: are strings adequately quoted? How about numbers, are without quotes?

This is already a good starting point.

src/c4/yml/emit.def.hpp
biojppm: :-)

Almost there...! Logic blocks now are ok, but the whole code is indented with respect to the namespace. Deindent it one level, as this project's style does not indent the namespace.

cburgard: should be done now

cburgard:
I can try giving it a go based on your instructions, but I don't know if I'll be able to pick up how your test suite syntax works. We'll see.

As for the "looking inside the JSON", I do have a conceptual question/problem here:
At the moment, since the nodes in the tree store everything as strings, all the elements of the JSON will be emitted as strings as well. As I don't think you plan on allowing different types of nodes (do you?), one would need to rely on "guessing" whether something was added to the tree as a string or as a number, which I don't find particularly elegant.
Thus, I would feel like the cleanest solution is to leave everything as strings, which I find fair seeing how javascript doesn't have strong typing anyway.
Do you think differently? Do you have a suggestion on how to cleanly distinguish string-type nodes from integer or float nodes?

biojppm:
First re creating the test. Add a test/basic_json.cpp file copying from basic.cpp and add the corresponding entry in the test/CMakeLists.txt file. 

Next, the test contents. For the emitting tests, [pick up from this](https://github.com/biojppm/rapidyaml/blob/db3df0f75941c733048d22fbc18d9ab5092e507a/test/basic.cpp#L1216), for example. Again, the test should parse a known tree (or create it by hand as in the tests above that example, but that's more tedious), then emitting, parsing the emitted string and emitting again. Then the strings must be equal. This tests the sanity of the emitter.

As for JSON types and quotes, indeed I am suggesting checking that in the absence of a yaml type tag, when we can tell that the values are numbers, they have no quotes. To be clear: if the scalar as a tag, we should respect its type. If there is no tag, we read through the string to try to figure out if it's a number.

This won't be perfect, but I'd say it is a better default than slapping quotes in every 0 or 1, which is the alternative. 

`c4::csubstr` has some handy utilities to create the predicate for checking that whether the string is a number; the following would be a starting point (won't work for every case but is ok for many cases):
```c++
bool is_number(c4::csubstr s)   // note, just improvising -- this is untested
{
    if(s.empty() || (s.first_non_empty_span().empty()) return false;
    return ((s.first_real_span() == s) || (s.first_int_span() == s) || (s.first_uint_span() == s);
}
```
Again, the cases where this is not ok may be dealt with by overriding with YAML type tags.

HTH!

cburgard:
I added code for the number handling now following your suggestion, it worked out nicely (except for a few missing braces).
However, I'm having a hard time getting the tests to actually run. I did not find any guidance in the documentation as to how to do this, and also couldn't understand how your CI is configured, so I just looked at the CMakeLists and tried this:
```cmake -DRYML_BUILD_TESTS=1 ..```
However, that only resulted in this:
```CMake Error at build/modules/c4fs/src/CMakeLists.txt:14 (c4fs_require_subproject):
  Unknown CMake command "c4fs_require_subproject".```
Is this a bug, or am I overlooking something obvious? How would I run the tests?

**Comment by biojppm** [comment]
@cburgard Thanks for persevering. What you did should work, but unfortunately some of the dependencies for testing are out of sync with ryml's current version of c4core/cmake. So this is indeed a bug, sorry.

But the good new are that this is already fixed in the dev branch, so if you rebase your changes on top of that branch, it should work; just verified that.

And don't forget to change the target branch of this PR to dev.

Again, thanks for persevering. This is indeed quite undocumented.

**Comment by cburgard** [comment]
I updated to the dev branch, which didn't fix it.
Then I moved to "master" in the c4core subproject, which didn't fix it.
Then I moved to "master" in the c4core/cmake subproject, which makes this appear:

    CMake Error at extern/c4core/cmake/c4Project.cmake:1219 (if):
    if given arguments:

    "STREQUAL" "Coverage"

    Unknown arguments specified 
     test/CMakeLists.txt:58 (c4_add_test)
     test/CMakeLists.txt:65 (ryml_add_test)

Could you point me to a combination of hashes of all of the projects involved here that actually manages to compile with the tests enabled?

**Comment by biojppm** [comment]
I was working on the install during the weekend and pushed to these branches, so this may have changed. Let me quickly confirm.

**Comment by biojppm** [comment]
OK, just confirmed this works on my end:

ryml: dev 2cb1abb
ryml/extern/c4core: master 2d694aa
ryml/extern/c4core/cmake: master b3b4eb6

ryml was not tracking the c4core master tip, so I pushed a commit to update c4core in ryml. If you rebase on the current dev, it should work. Don't forget to `git submodule update --recursive` so that the submodules stay in sync.

Again, sorry for the trouble. I'll think of a way to make this easier in the future.

**Comment by cburgard** [comment]
Still not working for me. I've checked the hashes, cleared the build - still no luck.
I've attached a log.
Which cmake version are you using? I'm using 3.14.7
[log.txt](https://github.com/biojppm/rapidyaml/files/3939256/log.txt)

**Comment by biojppm** [comment]
I see... CMAKE_BUILD_TYPE is not defined. I'll have to make sure this is set (as several points in the cmake code assume it exists and is not empty), but in the meantime, try to pass -DCMAKE_BUILD_TYPE=Debug (or Release) to cmake. 

I normally use a wrapper to cmake which sets it; and the unit tests also set it explicitly - that explains why I was not seeing that error.

**Comment by biojppm** [comment]
FYI, the install tests I added over the weekend are not yet succeeding, so expect to see CI failures from there.

**Comment by cburgard** [comment]
That works. Thanks a lot!

**Comment by cburgard** [comment]
Here's a weird one: 
Parsing of the emitted JSON generally fails, and I don't really understand why. Consider this:

    ERROR parsing yml: parse error
    line 1: '{"v2":"(10,11)","v3":"(100,101,102)","v4":"(1000,1001,1002,1003)"}' (sz=66)
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  (cols 6-67)

I'm pretty sure that ```{"v2":"(10,11)","v3":"(100,101,102)","v4":"(1000,1001,1002,1003)"}``` is valid JSON. I'm using `c4::yml::parse(csubstr&);` on it.
Any suggestion?

**Comment by biojppm** [comment]
There's a debug option (can't recall the name, but very verbose) which you can enable in the cmake file. Can you enable it and show me the log?

**Comment by biojppm** [comment]
Just looked at this again. It is not valid YAML; the colons are missing spaces after them. Although all the online validators say this is valid, they are wrong. [See this](https://stackoverflow.com/questions/42124227/why-does-the-yaml-spec-mandate-a-space-after-the-colon).

**Comment by cburgard** [comment]
Thanks for the help and explanations.

I just pushed a version that contains a first working JSON test, which is just a monkeypatched version of the basic YAML test.

File: src/c4/yml/emit.def.hpp [hunk 9]
```diff
@@ -268,4 +276,148 @@ void Emitter<Writer>::_write_scalar(csubstr s)
...
+    }
+    _do_visit(t, id);
+    if(t.is_stream(id))
+    {
+        _write("...\n");
+    }
+    return this->Writer::m_pos;
+}
+
+/** @todo this function is too complex. break it down into manageable pieces */
```
biojppm: this comment was for the original function. you can remove it here.

src/c4/yml/emit.def.hpp
```diff
@@ -268,4 +276,148 @@ void Emitter<Writer>::_write_scalar(csubstr s)
...
+    
+    _write_scalar(sc.scalar);
+}
+
+template<class Writer>
+void EmitterJSON<Writer>::_write_scalar(csubstr s)
+{
+    if(is_number(s))
+    {
+      this->Writer::_do_write(s);
```
biojppm: indentation

src/c4/yml/emit.hpp
```diff
@@ -241,6 +240,221 @@ CharOwningContainer emitrs(NodeRef const& n)
...
+struct as_json
+{
+    NodeRef node;
+    as_json(NodeRef const& n) : node(n) {}
+    as_json(Tree const& t) : node(t.rootref()) {}
+};
+template<class OStream>
+inline OStream& operator<< (OStream& s, as_json const& js)
+{
+    EmitterJSONOStream<OStream> em(*s.tree); // instantiate a json emitter instead
```
biojppm: this can't compile. `s` is a stream, and has no tree member. The tests should cover this.

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+
+    {
+        std::string tmp = "child5";
+        s.append_child() << tmp;
+        // now tmp can go safely out of scope, as it was
+        // serialized to the tree's internal string arena
+    }
+
+    emitrs_json(tree, &cmpbuf);
+    c4cmp=to_substr(cmpbuf);
```
biojppm: spacing

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+  - bar1
+  - bar2
+  - 33
+  - 44
+  - child5
+)";
+    EXPECT_EQ(cmpbuf2, exp);
+
+    // to serialize keys:
+    int k=66;
```
biojppm: spacing

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+    r.append_child() << key(k) << 7;
+
+    emitrs_json(tree, &cmpbuf);
+    c4cmp=to_substr(cmpbuf);
```
biojppm: spacing

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+    t["bar"] << 20;
+
+    m.clear();
+    t.rootref() >> m;
+
+    EXPECT_EQ(m["foo"], 10);
+    EXPECT_EQ(m["bar"], 20);
+}
+
+TEST(general, print_tree)
```
biojppm: please remove. this test does not touch any json logic, and is a pure duplicate of the original.

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+            y: b
+          -
+            z: c
+            u:
+)";
+    Tree t = parse(yaml);
+    print_tree(t); // to make sure this is covered too
+}
+
+TEST(general, lookup_path)
```
biojppm: no json, remove.

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+    EXPECT_EQ(dup[0].val().str, r[1][0].val().str);
+    EXPECT_EQ(dup[0].key().len, r[1][0].key().len);
+    EXPECT_EQ(dup[0].val().len, r[1][0].val().len);
+    EXPECT_EQ(dup[1].key().str, r[1][1].key().str);
+    EXPECT_EQ(dup[1].val().str, r[1][1].val().str);
+    EXPECT_EQ(dup[1].key().len, r[1][1].key().len);
+    EXPECT_EQ(dup[1].val().len, r[1][1].val().len);
+}
+
+TEST(NodeRef, intseq)
```
biojppm: no json, remove

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+    //printf("fonix"); print_tree(t); emit_json(r);
+    r[0].move(r[1]);
+    //printf("fonix"); print_tree(t); emit_json(r);
+    EXPECT_EQ(r[0].get(), m.get());
+    EXPECT_EQ(r[0].num_children(), map2.size());
+    EXPECT_EQ(r[1].get(), s.get());
+    EXPECT_EQ(r[1].num_children(), vec2.size());
+}
+
+TEST(NodeRef, 6_move_to_other_parent)
```
biojppm: no json, remove

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+    root.remove_child(8);
+    root.remove_child(6);
+    root.remove_child(4);
+    root.remove_child(2);
+    root.remove_child(0);
+
+    noderef_check_tree(root);
+}
+
+TEST(NodeRef, 5_move_in_same_parent)
```
biojppm: no json, remove

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+    EXPECT_EQ(root["b"]["aaa"].type(), KEYVAL);
+    EXPECT_EQ(root["b"]["aaa"].key_tag(), "!!str");
+    EXPECT_EQ(root["b"]["aaa"].key(), "aaa");
+    EXPECT_EQ(root["b"]["aaa"].val_tag(), "!!int");
+    EXPECT_EQ(root["b"]["aaa"].val(), "0");
+}
+
+//-----------------------------------------------------------------------------
+//-----------------------------------------------------------------------------
+//-----------------------------------------------------------------------------
```
biojppm: from here on, apparently all test are simple duplicates of the original, and test nothing specific to json emission.

Please remove.

test/basic_json.cpp
```diff
@@ -0,0 +1,1480 @@
...
+{
+    NodeInit n;
+    EXPECT_EQ(n.type, NOTYPE);
+    EXPECT_EQ(n.key.scalar, "");
+    EXPECT_EQ(n.key.tag, "");
+    EXPECT_EQ(n.val.scalar, "");
+    EXPECT_EQ(n.val.tag, "");
+}
+
+TEST(NodeInit, ctor__type_only)
```
biojppm: no json, remove

biojppm:
Other than the changes requested below, please review the copy, and only keep tests that specifically test json logic. I started by marking these, but there are too many. So please go through it again and remove pure duplicates.

src/c4/yml/emit.def.hpp
cburgard: done

src/c4/yml/emit.def.hpp
cburgard: done

src/c4/yml/emit.hpp
cburgard: fixed, test added

test/basic_json.cpp
cburgard: done

test/basic_json.cpp
cburgard: done

test/basic_json.cpp
cburgard: done

test/basic_json.cpp
cburgard: done

test/basic_json.cpp
cburgard: done

test/basic_json.cpp
cburgard: done

test/basic_json.cpp
cburgard: fixed

test/basic_json.cpp
cburgard: fixed

test/basic_json.cpp
cburgard: fixed, but this one is also there in the original `basic.cpp`, you might want to fix it there as well!

cburgard:
from my perspective it should all be done now. please re-review!

test/basic_json.cpp
```diff
@@ -192,6 +192,15 @@ TEST(general, map_to_root)
     EXPECT_EQ(m["bar"], 20);
 }
 
+TEST(general, json_stream_operator)
+{
+    const char *exp;
+    std::map<std::string, int> m({{"bar", 2}, {"foo", 1}});
+    Tree t;
+    t.rootref() << m;
+    std::cout << as_json(t) << std::endl;
```
biojppm: this is not a test. this is just coverage. a test requires EXPECTs and there are none here.

biojppm: send to a stringstream instead of cout, parse the output string and verify with calls to EXPECT that the trees compare equal.

cburgard: I would argue that it's still a test since it validates that the stream operator compiles successfully, but nevermind. I added some parsing and EXPECT calls :-)

test/basic_json.cpp
```diff
@@ -198,7 +198,13 @@ TEST(general, json_stream_operator)
     std::map<std::string, int> m({{"bar", 2}, {"foo", 1}});
     Tree t;
     t.rootref() << m;
-    std::cout << as_json(t) << std::endl;
+    std::stringstream ss;
+    ss << as_json(t);
+    Tree res = c4::yml::parse(to_csubstr(ss.str()));
```
biojppm: don't call `to_csubstr()` on a temporary! substr is a string view, and it makes you responsible for ensuring  that the lifetime of its memory fully contains the lifetime of the substr. 

So setting the csubstr from a temporary is UB and will likely crash. Instead, do this:
````c++
std::string result = ss.str();  // keep the memory in scope!
Tree res = c4::yml::parse(to_csubstr(result));
```

biojppm: Note that the test succeeded because std::string uses the small string optimization, and the produced yaml fits the small buffer. So there was no allocation, and hence no crash. But if the produced YAML is large enough, this will crash.

So I suggest using larger strings in the map and also a couple more nodes to make sure the string is not small and uses allocated memory.

cburgard: I added your suggested changes, however:
 * Even when using longer strings, the test still succeeded, perhaps because the memory was not re-used in the mean time and the read-only access thus still produced the correct result. But are you sure this is a real issue? I don't know enough about c++ to be exactly sure about how the lifetime of object behaves for nested function calls, but I would have expected this to work as long as it all happens within the same line. If you can point me to some explanation, I'd be happy to learn something new.
 * If you really want to avoid people using temporaries as arguments for the `substr`, is there any reason you don't do something like this here? https://stackoverflow.com/questions/11127807/prevent-pass-by-ref-of-temporary-object
But I'm getting ahead of myself I think, let's just take care of this PR :-)

test/basic_json.cpp
```diff
@@ -200,7 +200,8 @@ TEST(general, json_stream_operator)
     t.rootref() << m;
     std::stringstream ss;
     ss << as_json(t);
-    Tree res = c4::yml::parse(to_csubstr(ss.str()));
+    std::string result = ss.str();  // keep the memory in scope!
+    Tree res = c4::yml::parse(to_csubstr(result));
     m.clear();
     res.rootref() >> m;
     EXPECT_EQ(m["foo"], 1);
```
biojppm: Ha! This here is the problem... You are testing the original map: `m` is the map. You need to test the resulting tree...
So for instance, this
```c++
EXPECT_EQ(m["foo"], 1);
```
should be instead
```c++
EXPECT_EQ(res["foo"].val(), "1");  // note that "1" is and must be a string
```

This also explains why you were not seeing a crash (RAF, read after free) even after the string is no longer small, and hence using heap memory. That memory was not read from again **after** the parse() statement. Inside the parse() statement the temporary is still valid. As soon as the parse() returns, the temp is destroyed. If then you had read from it in the EXPECTs, you would have a crash.

You can also cause the RAF crash to occur inside parse if you do the following:
```c++
c4::csubstr s = ss.str();
// here, the temporary ss.str() was already destroyed so is pointing at stale memory
Tree res = c4::yml::parse(s);  // crash, RAF
```

test/basic_json.cpp
biojppm: That is a nice suggestion, thanks for that. I'm considering add a `to_csubstr(T &&)` overload just raising an error, ideally at compile time.

As for the error, read the comment to understand what's going on.

test/basic_json.cpp
biojppm: Well, I was wrong, sorry. Your line
```c++
res.rootref() >> m;
```
Does read from the memory. And indded, in my VS solution the test succeeds.

biojppm: Also tried with g++7 and clang10, valgrind and all the sanitizers, and it also succeeds. Weird.

I was using acquired wisdom, eg [this](https://stackoverflow.com/questions/4214153/lifetime-of-temporaries) or [this](https://stackoverflow.com/questions/4267676/do-temporary-objects-have-scope).

biojppm: Oh... I know.

Your call is `parse(to_csubstr(ss.str()));`. The `ss.str()` call creates a temporary. Indeed, the scope of this temporary is that of this expression, all the way to the `;`. So just like I said above, the temporary string will get destroyed AFTER parse() returns. Inside this expression, your call to `to_csubstr()` creates an _immutable_ string view. This view is valid inside the call to parse(). So parse will successfully finish.

But here is the detail that matters: parse() is given an immutable string. Note the ryml parser cannot parse directly from an immutable string. Because of YAML block scalars, block indentation and quote escaping behavior, ryml needs to modify the YAML source as it reads. So, when it is given the immutable string, ryml first copies into the tree's arena and parses from there. Then, all subsequent reads are viewing the tree's arena instead of the string's memory.

So that code is safe, because it is reading from memory in the tree's arena!

However.

If you slightly change this to use `to_substr(ss.str())` instead of `to_csubstr(ss.str())`, then you will indeed get the RAF from memory dangling out of the destroyed string. That's because now you've passed the mutable string view, which ryml will read from directly. So the parse will still succeed, but then as soon as you read from the tree via the `>>` operator, it will RAF.

So the recommendation still stands: when creating views to other objects, be careful with those other objects' lifetimes with the respect to the view object. Give names to those objects, and make sure their lifetime supersedes the view's lifetime. Views are extremely efficient and empowering, but they do have this mental overhead with the lifetime responsibility.

In this case, give a name to the temporary, and all the risk goes away, without any cost other than the extended lifetime of the temporary.

There. Sorry for misreading some of your code's details, and in particular this important mutability detail.

I'll have to review the API to make it clear to the caller that the immutable csubstr is first copied to the tree's arena. I fully admit that this situation is unsatisfactory.

biojppm: FYI, this is the test I adapted from your code:
```c++
TEST(general, json_stream_operator)
{
    std::map<std::string, int> out, m({{"bar", 2}, {"foo", 1}, {"foobar_barfoo:barfoo_foobar", 1001}, {"asdfjkl;", 42}, {"00000000000000000000000000000000000000000000000000000000000000", 1}});
    Tree t;
    t.rootref() << m;
    std::stringstream ss;
    ss << as_json(t);
    Tree res = c4::yml::parse(to_substr(ss.str()));
    EXPECT_EQ(res["foo"].val(), "1");
    EXPECT_EQ(res["bar"].val(), "2");
    EXPECT_EQ(res["foobar_barfoo:barfoo_foobar"].val(), "1001");
    EXPECT_EQ(res["asdfjkl;"].val(), "42");
    EXPECT_EQ(res["00000000000000000000000000000000000000000000000000000000000000"].val(), "1");
    res.rootref() >> out;
    EXPECT_EQ(out["foo"], 1);
    EXPECT_EQ(out["bar"], 2);
    EXPECT_EQ(out["foobar_barfoo:barfoo_foobar"], 1001);
    EXPECT_EQ(out["asdfjkl;"], 42);
    EXPECT_EQ(out["00000000000000000000000000000000000000000000000000000000000000"], 1);
}
```
Like I said, this fails spectacularly because of the RAF. To make it succeed, just do
```c++
TEST(general, json_stream_operator)
{
    std::map<std::string, int> out, m({{"bar", 2}, {"foo", 1}, {"foobar_barfoo:barfoo_foobar", 1001}, {"asdfjkl;", 42}, {"00000000000000000000000000000000000000000000000000000000000000", 1}});
    Tree t;
    t.rootref() << m;
    std::string str;
    {
        std::stringstream ss;
        ss << as_json(t);
        str = ss.str();
    }
    Tree res = c4::yml::parse(to_substr(str));
    EXPECT_EQ(res["foo"].val(), "1");
    EXPECT_EQ(res["bar"].val(), "2");
    EXPECT_EQ(res["foobar_barfoo:barfoo_foobar"].val(), "1001");
    EXPECT_EQ(res["asdfjkl;"].val(), "42");
    EXPECT_EQ(res["00000000000000000000000000000000000000000000000000000000000000"].val(), "1");
    res.rootref() >> out;
    EXPECT_EQ(out["foo"], 1);
    EXPECT_EQ(out["bar"], 2);
    EXPECT_EQ(out["foobar_barfoo:barfoo_foobar"], 1001);
    EXPECT_EQ(out["asdfjkl;"], 42);
    EXPECT_EQ(out["00000000000000000000000000000000000000000000000000000000000000"], 1);
}
```

cburgard:
Thanks for the deep investigation of this! This is all very interesting. I have updated the test according to your suggestion.
Anything else I should still change?

biojppm:
Nothing that jumps to the eye. Thanks for submitting, and kudos for persevering.

biojppm:
Your tests pass. Like I said, the fail is from the installer tests which are a WIP.

Squashing and merging. Thanks!

PR #77, opened 2020-08-16
Resolve VS 2019 warning C4296 -- Comparison always true comparing size_t >= 0

trprice:
Compare size_t parameters to > 0 instead of >= 0 to resolve VS 2019 warning C4296 that the >= 0 comparison will always be true. Given that size_t is defined as an unsigned this should be safe.

### DIFF
--- a/src/c4/yml/node.hpp
+++ b/src/c4/yml/node.hpp
@@ -612,7 +612,7 @@ public:
     inline void remove_child(size_t pos)
     {
         _C4RV();
-        RYML_ASSERT(pos = 0 && pos < num_children());
+        RYML_ASSERT(pos > 0 && pos < num_children());
         size_t child = m_tree->child(m_id, pos);
         RYML_ASSERT(child != NONE);
         m_tree->remove(child);
--- a/src/c4/yml/node.hpp
+++ b/src/c4/yml/node.hpp
@@ -612,7 +612,7 @@ public:
     inline void remove_child(size_t pos)
     {
         _C4RV();
-        RYML_ASSERT(pos >= 0 && pos < num_children());
+        RYML_ASSERT(pos = 0 && pos < num_children());
         size_t child = m_tree->child(m_id, pos);
         RYML_ASSERT(child != NONE);
         m_tree->remove(child);
--- a/src/c4/yml/tree.hpp
+++ b/src/c4/yml/tree.hpp
@@ -431,13 +431,13 @@ public:
     inline NodeData *get(size_t i)
     {
         if(i == NONE) return nullptr;
-        RYML_ASSERT(i = 0 && i < m_cap);
+        RYML_ASSERT(i > 0 && i < m_cap);
         return m_buf + i;
     }
     inline NodeData const *get(size_t i) const
     {
         if(i == NONE) return nullptr;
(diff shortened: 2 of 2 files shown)


trprice:
Sorry for taking several commits to catch all of the assignment mistakes. I changed them quickly and the assignments didn't show in my project build. Should've built rapidyaml by itself

trprice:
For some reason my local build was not displaying the error that the build in appveyor was showing

trprice:
Alright, the latest errors in convert.h weren't my changes.

biojppm:
Thanks for your merge request. I absolutely agree, these warnings must not be there and we need to find a way to suppress them. But I'm sorry to say the changes you are proposing are wrong.

You can either completely remove the `size_t pos >= 0` condition or suppress the warning using pragma directives. What you cannot do is remove the equal condition, ie change `pos >= 0` to `pos > 0` because 0 is a valid value in every one of those points.

The warning happens because size_t is always >= 0. However, I may in the future parameterize this code as a template of the index type, and there we cannot assume it will be an unsigned type. So that's why the assertions are there. The problem is that I do not see the warnings on my end. Can you share the Minimum Verifiable Example that triggers these warnings?

trprice:
Thank you for your response! That is fair that 0 may be a valid value in some of these cases.

I, unfortunately, fixed the symptom and did not extensively research where the functions are called and how the parameters are passed. If you have this knowledge at the forefront of your mind my question is: **Are these functions being called with size_t parameters?** In other words **If size_t is unsigned and must be 0 or greater why do we need to check for >= 0? It's guaranteed by the type**

By minimally verifiable example are you looking for minimum setup or minimum code?

And which compiler are you using? This error presented itself with Visual Studio 2019 v16.7.1

I will spin up a VM to check gcc and clang at some point today

biojppm:
Ha, thanks for pointing me towards those compiler flags.

That's very useful indeed; I've had some occasional reports of people seeing many warnings coming from the library but I was always scratching my head, as I did not know how it was they see it. I had clean compiles despite having my warning flags cranked up really high (BTW, above I said I was using /W3 but it was actually /W4).

I've picked up those compiler flags for warnings and integrated them in the project. A lot of warnings came out (mostly conversions to/from size_t), and I'm now fixing these.

biojppm:
Just noted I forgot to address this note of yours:

> As discussed in this thread I am ok disabling it in my project for this library with the note that, at present using size_t, the warning is technically correct. If and when the code is changed to pass in signed types then the assertion is totally warranted.

That's absolutely not the point, and you are not expected to do that. This library should not trigger any warning, period. End-users must not need to do anything at all related to suppressing warnings coming out of the library. As an end-user I do not like it when I have to do it, so I certainly do not want to do the same in my libraries.

Basically it boils down to this: if there is a warning, one of two things needs to happen in ryml's code:
- fix the code directly
- or if it is legitimate code (like the one here), pragma suppressions should be used *in the code*

trprice:
Thank you for the nice comments in the fix. I'm glad we could come to the point of getting things not only fixed but fixed in a way that is appropriate to the code base.

Given that this is a side project for me and I've got other stuff going I'm good with the changes happening in ryml

PR #102, opened 2021-01-09
Fixup python packaging

litghost:
This is my initial attempts at creating an installable python module for ryml.  Is this something that could be accepted?  How might this be changed to be better accepted?


api/ryml.i
```diff
@@ -117,6 +119,11 @@ void parse_substr(c4::substr s, c4::yml::Tree *t)
     c4::yml::parse(s, t);
 }
 
+std::string emit_std_string(const c4::yml::Tree &t)
+{
+    return c4::yml::emitrs<std::string>(t);
+}
```
biojppm, line 125: I see, the include is dictated by this. But then what is the specific pro for std::string in here? Would it not be possible to return a plain char* ? I'd rather stay away from std containers here, as they bring a lot of secondary linker questions with them. But if the arguments are strong enough, I may convinced otherwise.

api/ryml.i
```diff
@@ -200,6 +207,10 @@ def parse(buf, **kwargs):
     return _call_parse(parse_csubstr, buf, **kwargs)
 
 
+def emit(tree):
+    return emit_std_string(tree)
```
biojppm, line 211: There's also the question of idiom. Most of the idea in the current version is using memory views to avoid slow allocations, but this proposes allocating memory every call. Not entirely convinced here.

api/ryml.i
```diff
@@ -1,5 +1,6 @@
 
 %module ryml
+%include "std_string.i"
```
biojppm, line 3: see notes below

api/python/setup.py
```diff
@@ -1,17 +1,24 @@
...
+setup(name='rapidyaml',
+    version='1.0',
+    packages=['ryml'],
+    ext_modules=[ext],
+    cmdclass=dict(build_ext=BuildExtension),
+    author='Joao Paulo Magalhaes',
+    url='https://github.com/biojppm/rapidyaml',
+    description='This is a test',
+    license='MIT',
+    )
```
biojppm, line 24: Ideally, this should be generated from the properties in the main CMakeLists.txt. Not needed for now, but something to keep in mind.

biojppm (commented):
Thanks for the PR, it is really appreciated. However there are some questions of idiom that warrant some discussion. These can be settled quicker if we discuss them over call. Feel free to reach me in PM if you are up to it.

api/ryml.i
litghost: Returning `char*` won't work because the lifetime of the allocation would be tied a local within the function.  SWIG has wrappers for taking a std::string and converting to a python `str`, which is why `std::string` was used here.

Is there an allocating container in the `c4` library you would prefer over `std::string` to use with `emitrs`?  There are several viable solutions to use here, but `std::string` seemed like the most robust option at first glance.

api/ryml.i
litghost: What do you propose instead?  The user could pass a bytearray into the emit function, but then SWIG code would be needed to wrap the bytearray so that the C++ rapidyaml code could control the underlying allocation.  The maintenance burden in that code vs allocating one additional time per `emit` invocation seems like a reasonable trade-off.

api/ryml.i
biojppm: [char* should be fine with some SWIG plumbing taken care of](http://www.swig.org/Doc2.0/SWIGDocumentation.html#Library_nn11). Swig also provides [cstring.i](http://www.swig.org/Doc2.0/SWIGDocumentation.html#Library_nn12) which may be of use.

api/ryml.i
biojppm: I agree that the trade-off is generally reasonable; the point is that the user should be able to make the choice.

And considering that, I actually think your proposal of allocating is fine (just not `std::string`). We only have to provide another function, eg `emit_in_place()` which receives a `bytearray` that can be mutated. (This is not a request for correction, it's just me mapping how to enable power users with the capacity to avoid unnecessary allocations.)

api/ryml.i
litghost: > char* should be fine with some SWIG plumbing taken care of. 

This would require a `c4::yml::emitrs` method that operates on raw malloc'd `char*`s.  I don't believe that exists, does it?  

I also don't think that there is a difference in terms of allocation behavior between `%newobject` return `char *` and returing `std::string`.  I agree with the [other thread](https://github.com/biojppm/rapidyaml/pull/102#discussion_r555996812) discussion about pre-allocated output buffers, but that is a totally different case to returning a newly allocated buffer.

CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ c4_project(VERSION 0.1.0 STANDALONE
 option(RYML_DEFAULT_CALLBACKS "Enable ryml's default implementation of callbacks: allocate(), free(), error()" ON)
 option(RYML_BUILD_API "Enable API generation (python, etc)" OFF)
 option(RYML_DBG "Enable (very verbose) ryml debug prints." OFF)
+option(RYML_BUILD_TESTS "Enable build tests" OFF)
+option(RYML_BUILD_BENCHMARKS "Enable build benchmarks" OFF)
```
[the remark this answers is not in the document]
litghost: Removed

api/ryml.i
```diff
@@ -10,7 +11,8 @@
 // extension, inserting the module init code
 #define SWIG_FILE_WITH_INIT
 
-#include "ryml.hpp"
+#include <c4/yml/std/std.hpp>
```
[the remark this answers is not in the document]
litghost: Removed

api/ryml.i
litghost: I've changed this.

api/ryml.i
litghost: Removed

api/ryml.i
litghost: I've added the `emit_in_place` stuff.

api/ryml.i
biojppm: > This would require a c4::yml::emitrs method that operates on raw malloc'd char*s. I don't believe that exists, does it?

Indeed, and you figured correctly how to do it the same way. Kudos.

A clarification regarding the difference between char* and std::string. The question is not allocation behavior, but of linking requirements. If we stick to the c library then portability is much better. The complexity of sticking with malloc is not that much higher in terms of development. OTOH, exposing std::string in the public API (albeit implicitly) now brings into question std library link types, runtimes and so on. All of that just goes away by sticking to malloc(). In my view, that's a definite win.

api/ryml.i
```diff
@@ -200,6 +229,29 @@ def parse(buf, **kwargs):
...
+    if id is None:
+        id = tree.root_id()
+
+    (failed, expected_size) = emit_to_substr(tree, id, buf)
+    if failed:
+        raise IndexError("Output buffer has {} bytes, but emit required {} bytes".format(
+            len(buf), expected_size))
+
+    return memoryview(buf)[:expected_size]
+
```
biojppm, line 254: Nice. Thanks for going the extra mile. I was not asking you for `emit_in_place()`, but you did it. Much appreciated.

api/ryml.i
```diff
@@ -390,6 +442,9 @@ public:
     void set_key_ref   (size_t node, c4::csubstr ref   );
     void set_val_ref   (size_t node, c4::csubstr ref   );
 
+    void _set_key(size_t node, c4::csubstr key, int more_flags=0);
+    void _set_val(size_t node, c4::csubstr val, int more_flags=0);
```
biojppm, line 446: Why was this added? These are generally not to be called from client code, so I'm curious what the remaining API does not cover, but this covers instead.

litghost: Late key name binding basically.  `_set_key` is how `NodeRef` does this, and `NodeRef` is not currently exposed.  I haven't ended up needing `_set_val` at all.

biojppm:
@litghost please rebase; I'd like to see the tests succeed

litghost:
> @litghost please rebase; I'd like to see the tests succeed

Will do, no problem.

.github/workflows/ci.yml
```diff
@@ -542,12 +542,26 @@ jobs:
...
       - name: api-shared64-configure---------------------------------------------------
         run: source .github/setenv.sh && c4_cfg_test shared64
-      - {name: api-shared64-python-build, run: source .github/setenv.sh && c4_build_target shared64 ryml-api-python3}
-      - {name: api-shared64-python-test, run: source .github/setenv.sh && c4_build_target shared64 ryml-api-test-python3}
+      - name: api-shared64-python-build
+        run: |
+            python --version
+            python -mvenv env
+            source .github/setenv.sh
+            source env/bin/activate
```
biojppm, line 555: This is causing problems with the windows builds: https://github.com/biojppm/rapidyaml/pull/102/checks?check_run_id=1696849360#step:7:35

biojppm (approved):

.github/workflows/ci.yml
litghost: Fixed

biojppm:
@litghost The coverage and macosx failures in the tests are spurious. However, this one is meaningful: https://github.com/biojppm/rapidyaml/pull/102/checks?check_run_id=1712303885#step:8:97

litghost:
> @litghost The coverage and macosx failures in the tests are spurious. However, this one is meaningful: https://github.com/biojppm/rapidyaml/pull/102/checks?check_run_id=1712303885#step:8:97

Agreed.  I'm unclear on why the Windows test failed, I suspect some DLL hell is in play.  I'll try to replicate the failure this week!  Sorry for the long turn around.

litghost:
> > @litghost The coverage and macosx failures in the tests are spurious. However, this one is meaningful: https://github.com/biojppm/rapidyaml/pull/102/checks?check_run_id=1712303885#step:8:97
> 
> Agreed. I'm unclear on why the Windows test failed, I suspect some DLL hell is in play. I'll try to replicate the failure this week! Sorry for the long turn around.

Sorry for the long turn around.  I think this PR will now work on Windows, let's see what happens.

mithro:
Just FYI, I tried really hard to make it so the `setup.py`, `MANIFEST.in` and `pyproject.toml` could be in the `api/python` directory, but due to the way `pip` and `setuptools` interact, it is not possible for that to work and have packages which are publishable on PyPI. Once this pull request is merged, I can send a follow on pull request which will automatically publish packages (and wheels) onto PyPI every time to tag a release.

litghost:
Windows CI passes now!

```
     Creating library D:/a/rapidyaml/rapidyaml/build/shared64/api/python3/lib/Release/ryml.lib and object D:/a/rapidyaml/rapidyaml/build/shared64/api/python3/lib/Release/ryml.exp
  ryml-api-python3.vcxproj -> D:\a\rapidyaml\rapidyaml\build\shared64\api\python3\Release\_ryml.pyd
  Test project D:/a/rapidyaml/rapidyaml/build/shared64/api
      Start 1: ryml-api-test-python3-parse-run
  1/1 Test #1: ryml-api-test-python3-parse-run ...   Passed    0.23 sec
  
  100% tests passed, 0 tests failed out of 1
  
  Total Test time (real) =   0.32 sec
  Building Custom Rule D:/a/rapidyaml/rapidyaml/api/CMakeLists.txt
```

litghost:
@biojppm Is there something further you'd like to see changed?  Or is this good to merge?

biojppm:
Hey @litghost , apologies for the very late reply. Thanks again for all the hard work!

mithro:
@biojppm - Thanks for merging, we can clean up a bunch of stuff on our side now.

PR #104, opened 2021-01-12
Add val ref emission to emit.def.hpp.

litghost:


biojppm:
Thanks for the PR. This should be covered with a test. Starting with a [basic test](https://github.com/biojppm/rapidyaml/blob/13b70df5683ae9c9f105a43696cb9e02903ecea6/test/test_basic.cpp) is good, but even better would be one of the more thorough ones that test for idempotent parse/emit, for example [the anchor cases](https://github.com/biojppm/rapidyaml/blob/13b70df5683ae9c9f105a43696cb9e02903ecea6/test/test_simple_anchor.cpp#L99). To enable tests, simply configure cmake with `-DRYML_DEV=ON`.

litghost:
> Thanks for the PR. This should be covered with a test. Starting with a [basic test](https://github.com/biojppm/rapidyaml/blob/13b70df5683ae9c9f105a43696cb9e02903ecea6/test/test_basic.cpp) is good, but even better would be one of the more thorough ones that test for idempotent parse/emit, for example [the anchor cases](https://github.com/biojppm/rapidyaml/blob/13b70df5683ae9c9f105a43696cb9e02903ecea6/test/test_simple_anchor.cpp#L99). To enable tests, simply configure cmake with `-DRYML_DEV=ON`.

Agreed, I'll work on both some basic tests and the idempotent parse/emit tests.

litghost:
In adding tests, I think I see a potential issue with the current API that is ambiguous.

So the following test passes without the logic I added to this PR:
```
C("simple reference 1, unresolved",                                                                                                                                                                                                                                                       
R"(anchor_objects:                                                                                                                                                                                                                                                                        
  - &id001                                                                                                                                                                                                                                                                                
    name: id001                                                                                                                                                                                                                                                                           
  - &id002                                                                                                                                                                                                                                                                                
    name: id002                                                                                                                                                                                                                                                                           
  - name: id003                                                                                                                                                                                                                                                                           
  - &id004                                                                                                                                                                                                                                                                                
    name: id004                                                                                                                                                                                                                                                                           
references:                                                                                                                                                                                                                                                                               
  reference_key: *id001                                                                                                                                                                                                                                                                   
  reference_list:                                                                                                                                                                                                                                                                         
    - *id002                                                                                                                                                                                                                                                                              
    - *id004                                                                                                                                                                                                                                                                              
)",                                                                                                                                                                                                                                                                                       
L{                                                                                                                                                                                                                                                                                        
  N{"anchor_objects", L{                                                                                                                                                                                                                                                                  
   N(L{N("name", "id001")}, AR(VALANCH, "id001")),                                                                                                                                                                                                                                        
   N(L{N("name", "id002")}, AR(VALANCH, "id002")),                                                                                                                                                                                                                                        
   N(L{N("name", "id003")}),                                                                                                                                                                                                                                                              
   N(L{N("name", "id004")}, AR(VALANCH, "id004")),                                                                                                                                                                                                                                        
 }},                                                                                                                                                                                                                                                                                      
  N{"references", L{                                                                                                                                                                                                                                                                      
   N{"reference_key", "*id001", AR(VALREF, "id001")},                                                                                                                                                                                                                                     
   N{"reference_list", L{                                                                                                                                                                                                                                                                 
    N("*id002", AR(VALREF, "id002")),                                                                                                                                                                                                                                                     
    N("*id004", AR(VALREF, "id004")),                                                                                                                                                                                                                                                     
   }},                                                                                                                                                                                                                                                                                    
  }}                                                                                                                                                                                                                                                                                      
}                                                                                                                                                                                                                                                                                         
), 
```

but what this test is saying is that:
```
reference_list:
  - *id002
```

Parses as:
```
MAP(KEY = "reference_list", VAL=SEQ([VAL = "*id002", VALREF = "id002"]))
```

However I'd expect it to parse as:
```
MAP(KEY = "reference_list", VAL=SEQ([VALREF = "id002"]))
```

with the "VAL" bit set, etc.  What is the intended behavior of the VAL flag and `val` field in this case?

biojppm:
Ha, I have to look at this with some time; my mind is not fresh on this. Give me a day or two. Meanwhile, you can try to play around with `tree.resolve_references()`, which may be of help. There is the `RESOLVE_REFS` flag for the tests that forces `resolve_references()` to be called. Maybe you want to test without it?

biojppm:
id002 is a reference to a map containing one single entry, key=name/val=id002. What happens if you have a reference to a plain scalar? Should help narrow down where the problem is.

litghost:
> Ha, I have to look at this with some time; my mind is not fresh on this. Give me a day or two. Meanwhile, you can try to play around with `tree.resolve_references()`, which may be of help. There is the `RESOLVE_REFS` flag for the tests that forces `resolve_references()` to be called. Maybe you want to test without it?

In my use case, I actually don't want references resolved.  The test case I wrote doesn't use it.  The key question is, if a user is **constructing** a `c4::yml::Tree` and they want to emit:

```
anchor_objects:                                                                                                                                                                                                                                                                        
  - &id001                                                                                                                                                                                                                                                                                
    name: a_name
reference_list:
  - *id001
```
What do should they do.  I've created a smaller test case to show what I'm referring too:

https://github.com/biojppm/rapidyaml/pull/105/files#diff-64a5f4eaa5a7fc68b5d9e10c887491dc1c8233c57d75551d388f6ba2b10b2483R586-R619

biojppm (approved):

biojppm:
The proposed changes seem fine. But there was a glitch in the existing workflow action, that it did not trigger on PRs. I've fixed that on master. Please rebase, and then let's focus on having a test that verifiably fails your intended use case. Maybe that's the one from #104 after removing the line, but I'd still like to see it fail.

litghost:
> The proposed changes seem fine. But there was a glitch in the existing workflow action, that it did not trigger on PRs. I've fixed that on master. Please rebase, and then let's focus on having a test that verifiably fails your intended use case. Maybe that's the one from #104 after removing the line, but I'd still like to see it fail.

You can see the test fail over at https://github.com/biojppm/rapidyaml/pull/106

bm/CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ endfunction()
 
 c4_add_executable(ryml-bm-parse
     SOURCES bm_parse.cpp
-    LIBS ryml yaml yaml-cpp benchmark jsoncpp_lib nlohmann_json c4fs
+    LIBS ryml yaml yaml-cpp benchmark jsoncpp_static nlohmann_json c4fs
```
biojppm, line 67: Why did you need this? Never had the previous one fail. Maybe you configured with -DBUILD_SHARED_LIBS=OFF at the top level?

biojppm:
I am confused. These proposed changes are covered by the test added in #105 minus the line demonstrated to fail in #106, no? If that's the case, remove the line here, let's see it succeed. Otherwise, please clarify how this is covered in tests.

litghost:
> I am confused. These proposed changes are covered by the test added in #105 minus the line demonstrated to fail in #106, no? If that's the case, remove the line here, let's see it succeed.

I did remove the relevant line: https://github.com/biojppm/rapidyaml/pull/104/files#diff-64a5f4eaa5a7fc68b5d9e10c887491dc1c8233c57d75551d388f6ba2b10b2483L613 ?

bm/CMakeLists.txt
litghost: I didn't configure with `-DBUILD_SHARED_LIBS=OFF` explicitly, but that was indeed the problem.  I'll revert this change.  But I can replicate this issue, steps:
```
git clone https://github.com/biojppm/rapidyaml.git
cd rapidyaml/
mkdir build
git submodule update --init --recursive
cd build
cmake -DRYML_DEV=ON -GNinja ..
grep BUILD_SHARED_LIBS CMakeCache.txt 
```

Outputs:
```
BUILD_SHARED_LIBS:BOOL=OFF
```

at the end.  Invoking the CMake command as:
```
cmake -DRYML_DEV=ON -DBUILD_SHARED_LIBS=ON -GNinja ..
```
does avoid this issue.

litghost: I've removed this change.

biojppm:
> I did remove the relevant line

Indeed, I see it.

biojppm:
Thanks for the hard work.

PR #105, opened 2021-01-12
Add potentially failing test case for references.

litghost:


test/test_basic.cpp
```diff
@@ -560,6 +560,64 @@ TEST(serialize, std_string)
...
+    tree.set_val_anchor(anchor_map0, "id001");
+
+    auto anchor_elem0 = tree.append_child(anchor_map0);
+    tree.to_keyval(anchor_elem0, "name", "a_name");
+
+    auto ref_list_id = tree.append_child(root_id);
+    tree.to_seq(ref_list_id, "reference_list");
+
+    auto elem0_id = tree.append_child(ref_list_id);
+    tree.to_val(elem0_id, "*id001"); // FIXME: This line should not be required
```
biojppm, line 613: Good catch, indeed having to do this feels awkward.

litghost: This FIXME line is part of what #104 fixes.  When I wrote the idempotent test to check this, it fails because the parser effectively does this bit.

PR #106, opened 2021-01-12
DNM: Test that failing test case actually fails.

litghost:


biojppm:
Thanks for the verification

PR #107, opened 2021-01-13
Fix GH Actions YAML file.

litghost:
See https://github.com/biojppm/rapidyaml/actions/runs/481510848, which is still reporting a syntax error.


biojppm:
Thanks!

Issue #128, opened 2021-05-07
Segfault when referencing non-existent anchor


Gei0r:
`tree::resolve()` segfaults when the parsed yaml contains a reference to an anchor that does not exist.

Minimal example:
```c++
#include <ryml.hpp>

int main()
{
    const char filedata[] = "*";
    ryml::Tree tr = ryml::parse(filedata);
    tr.resolve();
}
```

Gei0r:
The problem is that `ReferenceResolver::_lookup()` runs into `C4_NEVER_REACH()` [here](https://github.com/biojppm/rapidyaml/blob/master/src/c4/yml/tree.cpp#L1209) when the anchor is not found.

I added a pull request that introduces proper error handling.

biojppm:
@Gei0r thanks for reporting, and thanks for submitting the fix.

PR #130, opened 2021-05-14
Fix segfault when accessing nonexistent anchor (#128)

Gei0r:
Fix for #128, including two small test cases.


biojppm:
@Gei0r Amazing, thanks for the fix. And sorry for the very late reply.

PR #132, opened 2021-05-15
Store quoted/unquoted information for parsed scalars; use it to distinguish references and strings starting with '*'

### LINKED ISSUE #129: Quoted strings starting with * are recognized as references
If a string starts with `*` and `tree::resolve()` is used, the string is recognized as a reference, even if it is quoted.

Example:
```c++
#include <ryml.hpp>

int main()
{
    const char filedata[] = R"(
a: &anchor "text"
b: "*anchor"
)";
    ryml::Tree tr = ryml::parse(filedata);
    tr.resolve();
    ryml::emit(tr);
}
```

Output:
```yaml
a: text
b: text
```
Expected Output:
```yaml
a: text
b: "*anchor"
```

Gei0r:
This PR adds new flags `KEYQUO` (key quoted) and `VALQUO` (value quoted) to parsed scalar nodes.

When resolving references, the quoting is taken into account, so that quoted scalars that happen to start with `*` are not wrongly "resolved".

This fixes #129.

As a side effect, this also allows distinguishing strings from numbers in a parsed document.

This PR also includes test cases.


biojppm (changes requested):
@Gei0r Amazing work, especially with getting to grips with the DSL from the tests. I'm requesting just a quick style change to mutate the variable in the parser.

Thanks for doing this, and sorry for taking so long to get back.

biojppm:
@Gei0r Also, can you rebase again?

biojppm:
@Gei0r I am merging now; I'll next add those style fixes on master. Again, thanks for your work!

Gei0r:
@biojppm I'm so happy you're back! Sorry for not reacting in due time to your questions.

PR #156, opened 2021-09-08
[fix] ensure ref-looking scalars are quoted

### LINKED ISSUE #151: Scalars beginning with * are emitted as invalid yaml
When setting a node value programmatically, it should be emitted with quotation if it starts with `*`, otherwise the emitted yaml is invalid.

Example:

```c++
#include <ryml.hpp>

int main()
{
    ryml::Tree tr;
    ryml::NodeRef root{tr.rootref()};
    root.set_type(ryml::MAP);

    ryml::NodeRef foo = root["foo"];
    foo = "*bar";

    ryml::emit(tr);   // foo: *bar     # INVALID, looks like a reference

    // explicitly set value to quoted - should happen automatically when value starts with *
    foo.set_type(foo.type() | ryml::VALQUO);

    ryml::emit(tr);   // foo: '*bar'   # OK
}
```

However, when *parsing* an unquoted value that starts with `*`, this should probably not be emitted with quotes, to preserve the reference when not calling `Tree::resolve()`.

biojppm:
fixes #151

### DIFF
--- a/test/test_simple_anchor.cpp
+++ b/test/test_simple_anchor.cpp
@@ -3,6 +3,46 @@
 namespace c4 {
 namespace yml {
 
+TEST(anchors, node_scalar_set_ref_when_empty)
+{
+    {
+        NodeScalar ns;
+        ns.set_ref_maybe_replacing_scalar("foo", /*has_scalar*/false);
+        EXPECT_EQ(ns.scalar, "foo");
+        EXPECT_EQ(ns.anchor, "foo");
+    }
+    {
+        NodeScalar ns;
+        ns.set_ref_maybe_replacing_scalar("*foo", /*has_scalar*/false);
+        EXPECT_EQ(ns.scalar, "*foo");
+        EXPECT_EQ(ns.anchor, "foo");
+    }
+}
+
+TEST(anchors, node_scalar_set_ref_when_non_empty)
+{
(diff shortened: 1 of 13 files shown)


test/test_simple_anchor.cpp
```diff
@@ -3,6 +3,410 @@
...
+        EXPECT_EQ(ns.anchor, "foo");
+        ns.set_ref_maybe_replacing_scalar("foo", /*has_scalar*/true);
+        EXPECT_EQ(ns.scalar, "*foo"); // keep the previous as it is well formed
+        EXPECT_EQ(ns.anchor, "foo");
+        ns.set_ref_maybe_replacing_scalar("bar", /*has_scalar*/true);
+        EXPECT_EQ(ns.scalar, "bar"); // replace previous as it is not well formed
+        EXPECT_EQ(ns.anchor, "bar");
+    }
+}
+
```
biojppm, line 45: @Gei0r I guess this PR is ready. The new behavior is consistent with what we previously discussed; but there are more nuances here, so take a look at the following tests to discover how it works, both from parsing and setting it programatically.

Gei0r: Looks really nice, I'll give it a thorough try tomorrow.

PR #166, opened 2021-11-01
Add missing node predicates in the Python API

### LINKED ISSUE #165: Some node predicate methods are missing in the Python API
In the Python API the following node predicate methods on `Tree` are missing (as I found out when trying to correctly parse `'True'` as a string)

* `is_anchor_or_ref()`
* `is_key_quoted()`
* `is_val_quoted()`
* `is_quoted()`

I'd be happy to try to create a PR adding those to the Python API.

simu:
This PR adds node predicates `is_anchor_or_ref()`, `is_quoted()`, `is_key_quoted()` and `is_val_quoted()` to the SWIG API definition for both `struct Tree` and `struct NodeType`.

Additionally, the PR changes the `SimpleHardCoded` Python API test case to have a quoted key and value respectively and adds assertions for the quoting-related predicates.

Fixes #165

### DIFF
--- a/api/python/Makefile
+++ b/api/python/Makefile
@@ -89,7 +89,7 @@ install: | $(ACTIVATE_SCRIPT)
 .PHONY: install
 
 test: | $(ACTIVATE_SCRIPT)
-	${ACTIVATE} pytest
+	${ACTIVATE} python -m nose ryml/tests
 
 .PHONY: test
 
--- a/api/python/ryml/tests/test_parse.py
+++ b/api/python/ryml/tests/test_parse.py
@@ -156,7 +156,7 @@ def check_tree_mod(ut, t):
 # -----------------------------------------------------------------------------
 class SimpleHardcoded:
 
-    yaml = "{HELLO: a, foo: b, bar: c, baz: d, seq: [0, 1, 2, 3]}"
+    yaml = "{'HELLO': a, foo: \"b\", bar: c, baz: d, seq: [0, 1, 2, 3]}"
 
     def check(self, ut, t):
         # some convenient shorthands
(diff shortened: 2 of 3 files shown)


api/python/requirements.txt
```diff
@@ -3,3 +3,4 @@ ninja
 pyyaml
 prettytable
 nose
+pytest
```
biojppm, line 6: Why this requirement?

simu: Without this requirement, I'm getting the following error when trying to run the Python API tests locally as shown below (it's entirely possible that I'm just not invoking the tests correctly):

```
simon@phoenix:~ $ cd ~/src/rapidyaml/api/python
simon@phoenix:~/src/rapidyaml/api/python $ make test
[ ... ]
Successfully installed rapidyaml-0.2.2.post16
Installed version: 0.2.2.post16
make[1]: Leaving directory '/home/simon/src/rapidyaml/api/python'
[[ -e venv/bin/activate ]] && source venv/bin/activate; pytest
/bin/bash: pytest: command not found
make: *** [Makefile:92: test] Error 127
simon@phoenix:~/src/rapidyaml/api/python $ 
```

simu: After looking at the GitHub workflows, I think the better option (to keep parity between CI and the Makefile) would be to apply the following change to `api/python/Makefile` instead of adding `pytest` to the requirements:

```diff
diff --git a/api/python/Makefile b/api/python/Makefile
index f5100b8..53db086 100644
--- a/api/python/Makefile
+++ b/api/python/Makefile
@@ -89,7 +89,7 @@ install: | $(ACTIVATE_SCRIPT)
 .PHONY: install
 
 test: | $(ACTIVATE_SCRIPT)
-       ${ACTIVATE} pytest
+       ${ACTIVATE} python -m nose ryml/tests
 
 .PHONY: test
```

biojppm: Agreed. Thanks for investigating.

simu: Pushed the suggested change instead of the additional dependency in the latest version

biojppm:
Thanks!

PR #169, opened 2021-11-25
[python] Added a draft of a transformer into native python objects.

KOLANICH:
Currently IDK how to identify a type of a value, so all the values are  strings.

### DIFF
--- a/api/python/ryml/__init__.py
+++ b/api/python/ryml/__init__.py
@@ -1,1 +1,90 @@
-from ryml.ryml import *
+import json
+from enum import IntFlag
+from typing import Dict, List, Optional, Union
+from warnings import warn
+
+from . import ryml
+from .ryml import *
+
+# pylint:disable=unused-argument
+
+NodeType = IntFlag("NodeType", {k: getattr(ryml, k) for k in ("NOTYPE", "VAL", "KEY", "MAP", "SEQ", "DOC", "STREAM", "KEYREF", "VALREF", "KEYANCH", "VALANCH", "KEYTAG", "VALTAG", "VALQUO", "KEYQUO",)})
+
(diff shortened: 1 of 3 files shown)


biojppm:
Thanks for the MR - it is definitely useful. Can you add unit tests?

KOLANICH:
Probably. Should I add them into an already existing file or should I create a new one?

biojppm:
You can create a new one, [in this folder](https://github.com/biojppm/rapidyaml/tree/master/api/python/ryml/tests)

KOLANICH:
@biojppm

biojppm:
What's up with [this error](https://github.com/biojppm/rapidyaml/runs/4355210622?check_suite_focus=true#step:7:228)?

KOLANICH:
It is extremily strange, especially given that my machine is 64-bit Ubuntu (for which the test pipeline also fails, it looks like it fails for all 64-bit machines, but for me the test works).

BTW what should I do to the `DOC` node? And is there any way to get a type (string, int, float, bool) of a value node?

KOLANICH:
@biojppm

KOLANICH:
`ruamel.yaml` `safe `parser is currently faster than ryml-based parser with my very rudimentary postprocessor.

2000 files:
`ruamel.yaml`: 23.7 s;
`ryaml.parse_to_python`: 20.5 s
`ryaml.parse_in_arena`: 0.5 s

1 file bigger and more complex than any of them (100 iterations)
`ruamel.yaml`: 18.0 s;
`ryaml.parse_to_python`: 22.3 s
`ryaml.parse_in_arena`: 0.45 s

So, if accelerate yaml parsing for python using `ryaml`, then the conversion should probably be done in native code.

PR #181, opened 2021-12-28
Added missing export for c4::yml::Tree class

aviktorov:


biojppm:
Thanks for the PR. Did this cause an actual problem to you? The CI extensively covers both shared and static libraries, but was flagging nothing with the missing export so I'm wondering. If you did see a problem, then something may be off in the CI, so that's why it is important to know.

aviktorov:
Hi there :) I'm happy to help!

It was causing link errors in my setup:
1. I made header-only rapidyaml library using amalgamate tool;
2. Then I put ryml_all.hpp to the public headers folder and put implementation inside one dll (during that stage I also had to add C4CORE_SHARED C4CORE_EXPORTS RYML_SHARED RYML_EXPORTS compiler definitions in order to have proper symbols exported). I guess amalgamate tool can help with that introducing another define which will set export definitions for both c4core and ryml;
3. After that, I tried to use first dll in another dll which contained specific parsing code and that caused linker errors because of missing exports.

I also found similar issues in debug builds with: c4::is_debugger_attached, c4::basic_substring<T> and some other symbols. Will do another pull request!

aviktorov:
Put another pull request to C4Core here: https://github.com/biojppm/c4core/pull/56

biojppm:
Thanks for the reply - it is interesting! The cmake scripts are [helping with the defines for the shared library](https://github.com/biojppm/cmake/blob/9416f297430bc358eea48f8cf65aa0d602029629/c4Project.cmake#L1854) (BTW, I guess you're using windows?). I will try to reproduce in the CI from the amalgamated header - I see now that no CI test was using amalgamated as a shared library.

The amalgamated header was merged just today, so that's still a bit rough around the edges. I opened a new issue to track the missing exports. Hopefully no more major issues are outstanding.

PR #183, opened 2021-12-29
[wip] fix exports in single header

### LINKED ISSUE #182: Missing exports for symbols in amalgamated header
_Originally posted by @aviktorov in https://github.com/biojppm/rapidyaml/issues/181#issuecomment-1002268297_

Missing exports are causing link errors:
1. I made header-only rapidyaml library using amalgamate tool;
2. Then I put ryml_all.hpp to the public headers folder and put implementation inside one dll (during that stage I also had to add C4CORE_SHARED C4CORE_EXPORTS RYML_SHARED RYML_EXPORTS compiler definitions in order to have proper symbols exported). I guess amalgamate tool can help with that introducing another define which will set export definitions for both c4core and ryml;
3. After that, I tried to use first dll in another dll which contained specific parsing code and that caused linker errors because of missing exports.

biojppm:
fixes #182

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -529,6 +529,8 @@ jobs:
       - {name: show info, run: source .github/setenv.sh && c4_show_info }
       #
       - {name: singleheader, run: cd samples/singleheader && ./run.sh $BT }
+      - {name: singleheaderlib-static, run: cd samples/singleheaderlib && ./run_static.sh $BT }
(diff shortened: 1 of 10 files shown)


biojppm:
@aviktorov this works for me as it is. can you try now? All you need to do is `#define RYML_SHARED`, nothing else should be needed.

biojppm:
Merging now. Feel free to reopen the bug if it is not right yet.

Issue #37, opened 2020-02-07
emit formatting


biojppm:
Eg, #29 or #30 . Some thoughts:
* implement custom `EmitFormatter`s:
   * block (the current emitter)
   * single-line flow
   * multi-line flow
   * semi-intelligent: use a general emitter, using heuristics such as counting columns and breaking and selecting block/flow styles as appropriate (eg, nested containers)
* implement per-node format flags? i.e., add flags forcing format selection?
* add type-based implicit tagging? e.g. implicit tag bools so that they show as "true" rather than "1"

leoetlino:
Sounds good to me. Per-node flags would be very useful, since it would let users override the style and get exactly what they want. I assume there'd be three flags, block/flow/auto, with block being set by default?

Another thing that would be nice to have is the ability to enable wrapping (by setting a maximum line width). With all these additions ryml would be on par with libyaml for formatting options.

>add type-based implicit tagging? e.g. implicit tag bools so that they show as "true" rather than "1"

Could this not be achieved just by changing the format overload for bool to give "true"/"false" rather than 0/1, or am I missing something?

rysavyjan:
I'm using Rapid YAML for merging YAML output from clang-tidy into [replacements file](https://devblogs.microsoft.com/cppblog/exploring-clang-tooling-using-build-tools-with-clang-tidy/).

Now is following input
```
ReplacementText: "#ifndef E:\\SOURCE\\FINE\\COMMON\\GRAPHICSUTILS_H\n#define E:\\SOURCE\\FINE\\COMMON\\GRAPHICSUTILS_H\n\n"
```

emitted as

```
ReplacementText: |+
  #ifndef E:\SOURCE\FINE\COMMON\GRAPHICSUTILS_H
  #define E:\SOURCE\FINE\COMMON\GRAPHICSUTILS_H

```

It would be great to have option to output value in the original format.
I assume it is not possible with current ryml version? (I will probably use some dirty hack for my use case.)

biojppm:
@rysavyjan sorry for the late reply, somehow I missed your post.

You are correct; at the moment ryml does not implement this. This issue is meant to track progress in formatting, but I've not been able to do anything since I opened it. I also won't be able to come back to this in the next month or two.

captain-yoshi:
Feedback and ideas from using `_WIP_STYLE_FLOW_SL`:
```cpp
[...]

PR #191, opened 2022-01-17
[wip] emit styles

biojppm:
re #37


captain-yoshi:
Just giving feedback for ` _WIP_STYLE_FLOW_SL`:
```cpp
_WIP_STYLE_FLOW_SL = c4bit(14), ///< mark container with single-line flow format (seqs as '[val1,val2], maps as '{key: val, key2: val2}')

// Is this an erro ? There should be a space between values like for the map. '[val1, val2]'
```

You could even have these 2 styles :
```yaml
# _WIP_STYLE_FLOW_SL
[val1, val2]

# _WIP_STYLE_FLOW_SL_COMPACT
[val1,val2]
```

PR #199, opened 2022-01-22
Fixed typo

PixelParas:
Before: 
```... so, it don't see it as pratical```

After:
```... so, I don't see it as practical```


biojppm:
thanks!

PR #216, opened 2022-02-13
[fix] ensure keys starting with `%` are quoted to prevent them from being treated as a directive

xTVaser:
Hey, ran into this small edge-case when working with this library.  Consider the initial YAML file:
```yaml
'%ROOT':
  timeCreated: 1642559912
  timeModified: 1642553437
somethingElse:
  timeCreated: 1642559912
  timeModified: 1642553437
```
After parsing and re-emitting this:
```cpp
std::optional<std::vector<u8>> buf(Host::ReadResourceFile("test.yaml"));
const ryml::substr view = c4::basic_substring<char>(reinterpret_cast<char*>(buf->data()), buf->size());
ryml::Tree tree = ryml::parse(view);

const std::string full_filename(Path::CombineStdString(EmuFolders::Resources, "test-modified.yaml"));
FILE* modifiedFile = fopen(full_filename.c_str(), "w");
ryml::emit(tree, tree.root_id(), modifiedFile);
fclose(modifiedFile);
```
The quotes are lost:
```yaml
%ROOT:
  timeCreated: 1642559912
  timeModified: 1642553437
somethingElse:
  timeCreated: 1642559912
  timeModified: 1642553437
```

The `%ROOT` key will now be treated as a preprocessor directive (i assume), which means the file will unexpectedly be handled differently and modified.  Parsing the above and re-emitting results in the following:
```yaml
timeCreated: 1642559912
timeModified: 1642553437
somethingElse:
  timeCreated: 1642559912
  timeModified: 1642553437
```

Which completely changes the schema.  Ensuring keys that begin with `%` are quoted seems to prevent this edge-case.

Relevant spec docs - https://yaml.org/spec/1.2.2/#example-directive-indicator

### DIFF
--- a/src/c4/yml/emit.def.hpp
+++ b/src/c4/yml/emit.def.hpp
@@ -827,8 +827,8 @@ void Emitter<Writer>::_write_scalar(csubstr s, bool was_quoted)
                 // has leading whitespace
                 s.begins_with_any(" \n\t\r")
                 ||
-                // looks like reference or anchor
-                s.begins_with_any("*&")
+                // looks like reference or anchor or would be treated as a directive
+                s.begins_with_any("*&%")
                 ||
                 s.begins_with("<<")
                 ||


biojppm:
Hi @xTVaser , thanks for the fix.

I was able to reproduce the problem, but not with your approach. Are you based in the latest master? (Also, I noticed you are using `ryml::parse()` which has been deprecated in favor of either `parse_in_place()` or `parse_in_arena()`; see the changelog for the explanation). 

I ask that because since 0.3.0 (I think) the input quotes are preserved in a roundtrip. That is, if a scalar is quoted in the source YAML, it will be quoted when emitting as well. So this test succeeds (and note it checks over several parse/emit cycles):
```c++
TEST(emit, percent_quotes_are_preserved)
{
    csubstr yaml = R"(
'%ROOT': '%VAL'
'%ROOT2':
  - '%VAL'
  - '%VAL'
)";
    test_check_emit_check(yaml, [](Tree const &t){
        ASSERT_TRUE(t.rootref().is_map());
        ASSERT_TRUE(t.rootref().has_child("%ROOT"));
        ASSERT_TRUE(t.rootref().has_child("%ROOT2"));
        ASSERT_EQ(t["%ROOT2"].num_children(), 2u);
        EXPECT_TRUE(t["%ROOT"].is_key_quoted());
        EXPECT_TRUE(t["%ROOT"].is_val_quoted());
        EXPECT_TRUE(t["%ROOT2"].is_key_quoted());
        EXPECT_TRUE(t["%ROOT2"][0].is_val_quoted());
        EXPECT_TRUE(t["%ROOT2"][1].is_val_quoted());
    });
}
```
So the current master does not have the problem as you described.

However, if we set the scalar programmatically, then it is emitted without quotes unless we explicitly set it to quoted. So this fails indeed:
```c++
TEST(emit, percent_is_quoted)
{
    Tree ti = parse_in_arena("{}");
    ASSERT_TRUE(ti.rootref().is_map());
    ti["%ROOT"] = "%VAL";
    ti["%ROOT2"] |= SEQ;
    ti["%ROOT2"][0] = "%VAL";
    ti["%ROOT2"][1] = "%VAL";
    std::string yaml = emitrs<std::string>(ti);
    test_check_emit_check(to_csubstr(yaml), [](Tree const &t){
        ASSERT_TRUE(t.rootref().is_map());
        ASSERT_TRUE(t.rootref().has_child("%ROOT"));
        ASSERT_TRUE(t.rootref().has_child("%ROOT2"));
        ASSERT_EQ(t["%ROOT2"].num_children(), 2u);
        EXPECT_TRUE(t["%ROOT"].is_key_quoted());
        EXPECT_TRUE(t["%ROOT"].is_val_quoted());
        EXPECT_TRUE(t["%ROOT2"].is_key_quoted());
        EXPECT_TRUE(t["%ROOT2"][0].is_val_quoted());
        EXPECT_TRUE(t["%ROOT2"][1].is_val_quoted());
    });
}
```
And your fix successfully addresses this problem.

biojppm:
@xTVaser the test failures are spurious, and your fix demonstrably works. Merging now.

Thanks for the PR!

xTVaser:
Sorry for the delayed response, thanks for the merge.  Shame on me for not pulling down the latest changes 😄 , I guess i figured it was pretty obscure and probably wasn't fixed -- glad to hear this still partially fixed something though!

Thanks for pointing out the `parse` deprecation as well, I'll switch our stuff over 👍

PR #247, opened 2022-04-13
modify sample_create_trees()

1099255210:
In `quickstart.cpp`, due to line 1749, the sample wants to show how to use `<<` to avoid lifetime dependency. Thus, a `std::string a_drop` is defined, but after this `a_drop` is not really used.


biojppm:
Indeed, thanks for the fix.

PR #249, opened 2022-04-15
[fix] specify python3 command for the amalgamate tool

captain-yoshi:
In Ubuntu 20.04 the default `python` [command](https://github.com/biojppm/rapidyaml/blob/master/samples/singleheader/amalgamate.cmake#L12-L13) is associated with python 2.7.X. This PR changes the amalgamate tool command to use `python3` because of the incompability with python2.

Alternatives:
- Change `f-Strings` to `str.format()` ... 
- Add `#!/usr/bin/env python3` in file and call without python command (also make script executable). Would be best for portability ? E.g I suspect python and python3 commands to not work in windows...

I can remove the change in the doc if you want.

### DIFF
--- a/samples/singleheader/amalgamate.cmake
+++ b/samples/singleheader/amalgamate.cmake
@@ -1,3 +1,4 @@
+find_package(Python3 COMPONENTS Interpreter)
 
 # amalgamate ryml to get the single header
 function(amalgamate_ryml header_dir header_file)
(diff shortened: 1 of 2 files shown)


samples/singleheader/amalgamate.cmake
```diff
@@ -9,8 +9,8 @@ function(amalgamate_ryml header_dir header_file)
         LIST_DIRECTORIES FALSE
         CONFIGURE_DEPENDS "${rymldir}/src")
     add_custom_command(OUTPUT "${singleheader}"
-        COMMAND python "${amscript}" "${singleheader}"
-        COMMENT "python ${amscript} ${singleheader}"
+        COMMAND python3 "${amscript}" "${singleheader}"
+        COMMENT "python3 ${amscript} ${singleheader}"
```
biojppm, line 13: The proper fix here is by calling [find_package(Python3 ...)](https://cmake.org/cmake/help/latest/module/FindPython3.html) and then using `${Python3_EXECUTABLE}` instead of `python` or `python3`.

captain-yoshi: Much cleaner!

captain-yoshi:
It works on my end with your proposed fix. The check suite seems to be queued.

biojppm:
Thanks!

captain-yoshi:
I received an email mentioning that some [tests](https://github.com/biojppm/rapidyaml/actions/runs/2172825904) were failing. Don't know if it's important.

biojppm:
Thanks for the heads up.

PR #267, opened 2022-06-27
Add ConstNodeRef to enforce const correctness

### LINKED ISSUE #266: const on return types
While looking quickly at the code I often see things like:

```cpp
NodeRef operator[] (csubstr k)
{
    // [...]
    NodeRef r = ...;
    return r;
}

NodeRef const operator[] (csubstr k) const
{
    // [...]
    NodeRef const r = ...;
    return r;
}
```
I might be missing something but I see two problems with this pattern:
 1. having a const value as the return type makes no sense since the returned value can be assigned to a non-const one by making a copy
 2. returning a const value prevents moves which in turn prevents NRVO

See the [relevant clang-tidy check](https://releases.llvm.org/11.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-const-return-type.html)

If you want constness propagated to returned values it must be encoded in the type. For instance you could have a `NodeConstRef` with only the const-qualified member functions of `NodeRef` and use that as a return type for such const-qualified functions.

biojppm:
fixes #266


BenjaminNavarro:
Nice, didn't thought this was going to be taken care of so quickly.

To minimize code duplication I guess you could have either:
 1. `NodeRef` inherits from `ConstNodeRef` and adds the state modifying functions
 2. Have two separate classes, e.g `NodeReadOperations` and `NodeWriteOperations` and have `ConstNodeRef` inherit from the first and `NodeRef` from both.

I think 2. is what is used in Eigen, but that might require some CRTP to pull off. 1. is easy but some people don't like to use inheritance for these things;

Issue #292, opened 2022-08-10
Syntax error when emitting to JSON from a YAML versioning scheme value


captain-yoshi:
Emitting a JSON file from a YAML with versioning value will give a SyntaxError when parsed in JSON.

YAML
``` yaml
james: 0.0.7
```
JSON
``` json
{"james": 0.0.7}
```
The versioning conversion to a double should not have passed (just a hypothesis).

captain-yoshi:
@jppm I'm getting an unquoted versioning for `1.1.9` when converted to json.

Working versioning:
- `0.6.1`
- `0.0.0`

biojppm:
That's odd. It succeeds with this:
```c++

TEST(emit_json, issue292)
{
    EXPECT_FALSE(csubstr("0.1.0").is_number());
    Tree t = parse_in_arena("{}");
    t["james"] = "1.2.3";
    EXPECT_EQ(emitrs_json<std::string>(t), "{\"james\": \"1.2.3\"}");
    t["james"] = "1.1.9";
    EXPECT_EQ(emitrs_json<std::string>(t), "{\"james\": \"1.1.9\"}");
    t["james"] = "0.6.1";
    EXPECT_EQ(emitrs_json<std::string>(t), "{\"james\": \"0.6.1\"}");
    t["james"] = "0.0.0";
    EXPECT_EQ(emitrs_json<std::string>(t), "{\"james\": \"0.0.0\"}");
}
```
Are you sure you have the latest version?

captain-yoshi:
Latest version with rapidyaml. Does fetching rapidyaml through git also fetches the submodules ?

captain-yoshi:
Forgot to update with `git submodule update --init --recursive` ... and made a PR to update this command in the docs... Oh well !

PR #302, opened 2022-09-07
Adds a unit test to parse 3 digits versions

ericcitaire:
See https://github.com/google/jsonnet/issues/1014

### DIFF
--- a/test/test_case.cpp
+++ b/test/test_case.cpp
@@ -854,12 +854,16 @@ CaseData* get_data(csubstr name)
             replace_all("\r", "", c->src, &tmp);
             cd->unix_style.src_buf.assign(tmp.begin(), tmp.end());
             cd->unix_style.src = to_substr(cd->unix_style.src_buf);
+            cd->unix_style_json.src_buf.assign(tmp.begin(), tmp.end());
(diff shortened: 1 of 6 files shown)


biojppm:
tangentially related: #292

biojppm:
@ericcitaire I force-pushed a fix to your commit, and elaborated the tests. I don't know what is the particular problem you are having (the linked bug does not provide any context and is unclear for me), but the tests now do a successful roundtrip of multi-dot version strings, both in YAML and JSON. If the CI passes, this should be evidence that the problem is neither in ryml's parsing nor emitting.

biojppm:
Indeed the tests succeed (the 9 failures are unrelated). Merging now. Thanks!

ericcitaire:
> @ericcitaire hard to see from [google/jsonnet#1014](https://github.com/google/jsonnet/issues/1014). Is it a duplicate of #292 (already fixed in master)?

Indeed, it seems to be a duplicate of #292. Sorry, I missed it.

So I guess [google/jsonnet](https://github.com/google/jsonnet) just have to rebuild with a recent version of rapidyaml.

Thanks !

PR #310, opened 2022-09-27
[fix] emitted scalars containing @ should be quoted

### LINKED ISSUE #309: Indicator character emitted as plain scalar unexpectedly
When doing something like this:
```cpp
ryml::Tree tree;
ryml::NodeRef wroot = tree.rootref();
wroot |= ryml::MAP;
wroot["name"] << "@test";
ryml::emit_yaml(wroot, stdout)
```
The output is 
```
name: @test
```
while it is expected to be
```
name: '@test'
```
since "reserved indicators can't start a plain scalar."

I'm aware that ryml tends to be more permissive but in this case the generated output
cannot be parsed by other standard compliant parsers.

Reference: https://yaml.org/spec/1.2.2/#57-escaped-characters, example 5.10

biojppm:
fixes #309

### DIFF
--- a/changelog/0.5.0.md
+++ b/changelog/0.5.0.md
@@ -99,6 +99,7 @@
   Tree tree2 = parse_in_arena(to_csubstr(s));
   EXPECT_EQ(tree2["s"].val(), tree["s"].val());
   ```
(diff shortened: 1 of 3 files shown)


perlpunk (commented):
not sure if it is already covered, I don't see the backtick in the list, it also has to be quoted

biojppm:
@perlpunk thanks for the note. indeed backtick was also missing.

PR #328, opened 2022-11-18
[fix] eager assertions prevented moving nodes to the first position

### LINKED ISSUE #324: Move node to first position
Hi,

is it possible to use the node ``move`` operation in order to move a node along it's siblings to the first place? Right now, there is only an ``after`` option. Using a NodeRef {tree, c4::yml::NONE } as after, doesn't seem to work, as it is not a sibling.

Thanks for this library!

biojppm:
fixes #324


biojppm:
The failures in the CI runs are unrelated to this changeset, and from use of a python package for plotting benchmark results. This is good to merge.

PR #338, opened 2022-12-02
fix empty line detection

### LINKED ISSUE #337: Potentially incorrect empty line detection in `Parse::_filter_block_scalar`
In this method if `chomp != CHOMP_KEEP` is true, then `s.trim(" \n\r\t")` is executed.
Obviously this statement shall filter empty lines, but if I read the YAML 1.2 standard correctly, then an empty line in block scalars cannot contain any `\t`-character.

That seems to be the reason, why rapidyaml fails [this test case](https://github.com/yaml/yaml-test-suite/blob/main/src/Y79Y.yaml#L16).

NaN-git:
Empty lines in block scalars cannot contain a `\t`-character.

All non-failing YAML parsers on YAML playground do not recognize the `\t`-character as part of an empty line, too.

Fixes #337


biojppm:
@NaN-git thanks! This is really appreciated.

PR #381, opened 2023-06-29
Do not mark visit functions as pure

tbleher:
Marking a function as pure implies that it does not have any observable side effects (only the return value is used). So if the compiler detects that the return value of a pure function is not used, it may eliminate the call to the pure function.

This would be wrong, however, if the passed in Visitor has side effects which are needed by the program.
Fix this by not marking the visit functions as pure.


biojppm:
Thanks!

PR #383, opened 2023-07-17
Update natvis to include a few tweaks

### LINKED ISSUE #380: Natvis support is missing some ryml types
Natvis is an XML-based description format for types in the debugger.  It is supported by windbg.exe and the Visual Studio debugger, at the very least.  Visualizations in the debugger are especially helpful for things that are difficult to decode in the debugger (e.g. indexes into a big block of memory are difficult whereas a simple integer is easy).  The documentation on natvis is here: https://learn.microsoft.com/en-us/visualstudio/debugger/create-custom-views-of-native-objects?view=vs-2022.

I have been trying to add new usage of rapidyaml to a project that I'm working on and even in debug mode I have not been able to make sense of the ryml types in the debugger.

I had the same problem with rapidjson.  However, I just looked and they support natvis.  Applying their natvis file to my system was a night and day difference.  Their natvis file lives here: https://github.com/Tencent/rapidjson/blob/master/contrib/natvis/rapidjson.natvis.

dmachaj:
## Why is this change being made?
I set up the natvis in a new project that is using rapidyml and it was missing the primary type that we are using (`ConstNodeRef`).  Additionally I discovered that the children synthetic visualization for NodeRef/ConstNodeRef was not being parsed correctly.  The order of operations on the bitwise operator was causing the condition to always evaluate to false.

## Briefly summarize what changed
* Duplicate the `NodeRef` visualization as `ConstNodeRef`.  ConstNodeRef does not have an m_seed field so it cannot be exactly identical or else it fails to apply.
* Tweak `NodeData::m_parent` to special-case `NONE` instead of printing uint64(-1).
* Fix the [children] expansion in NodeRef/ConstNodeRef.
* Expand `elm` to `element`.  (This one is arguable; I'm tempted to expand them all out so they are easier to understand without abbreviation).

I also tried to get `NodeData` to have a synthetic child with the expansion of the sub-elements (e.g. `KEYMAP` can print the keys in the map).  Unfortunately this does not seem to be possible because the `NodeData` has the raw indexes but no pointer to the memory allocation.

## How was this change tested?
I have a ~10 line program using rapidyml along with an OpenAPI.yml file from a public sample.

Closes #380

### DIFF
--- a/src/ryml.natvis
+++ b/src/ryml.natvis
@@ -71,7 +71,8 @@ See also:
       <Item Name="val ref" Condition="(m_type.type &amp; c4::yml::VALREF) != 0">m_val.anchor</Item>
       <Item Name="key anchor" Condition="(m_type.type &amp; c4::yml::KEYANCH) != 0">m_key.anchor</Item>
       <Item Name="val anchor" Condition="(m_type.type &amp; c4::yml::VALANCH) != 0">m_val.anchor</Item>
-      <Item Name="parent">m_parent</Item>
+      <Item Name="parent" Condition="m_parent == c4::yml::NONE">NONE</Item>
+      <Item Name="parent" Condition="m_parent != c4::yml::NONE">m_parent</Item>
       <Item Name="first child"  Condition="m_first_child != c4::yml::NONE">m_first_child</Item>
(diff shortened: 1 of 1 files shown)


biojppm:
LGTM. Thanks!

Issue #391, opened 2023-11-02
Support for Python 3.12


mboutet:
I think it's only a matter of making the wheel available for Python 3.12 on PyPI.

On `Python 3.11.3`, I'm able to install it via `pip install rapidyaml`, but on `Python 3.12.0`, it tries to build the wheel and fails.

I also tried with Poetry, but it complains that it's unable to find an installation candidate:

```
  • Installing rapidyaml (0.5.0): Failed

  RuntimeError

  Unable to find installation candidates for rapidyaml (0.5.0)
```

Let me know if you need more information.

Thank you

PR #392, opened 2023-11-03
chore: Add support for Python 3.12

mboutet:
Really not sure that it's the only thing I need to do to support Python 3.12 😛 

Fixes https://github.com/biojppm/rapidyaml/issues/391

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -205,6 +205,8 @@ jobs:
       matrix:
         config:
           # the 3-digit versions NEED to be quoted to prevent the version being read as float. (!)
+          - {pythonv: '3.12', cibw_pyv: 311, cibw_arch: x86_64, cibw_platform: manylinux_x86_64, osname: linux, os: ubuntu-20.04}
+          - {pythonv: '3.12', cibw_pyv: 311, cibw_arch: i686  , cibw_platform: manylinux_i686  , osname: linux, os: ubuntu-20.04}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -245,7 +247,7 @@ jobs:
       # use fetch-depth to ensure all tags are fetched
       - {name: checkout, uses: actions/checkout@v3, with: {submodules: recursive, fetch-depth: 0}}
       - name: create wheel
-        uses: joerick/cibuildwheel@v2.9.0
+        uses: pypa/cibuildwheel@v2.16.2
```
mboutet, line 250: https://github.com/joerick/cibuildwheel was relocated to https://github.com/pypa/cibuildwheel and I also bumped to the latest version which added support for Python 3.12.

mboutet:
@biojppm, sorry for the ping, but any chance you can approve the workflows so they can run?

mboutet:
I'm not sure if the failures in the CI are due to my changes. The failure for the `samples` workflow appears to be caused by the workflow trying to checkout commit https://github.com/biojppm/rapidyaml/pull/392/commits/c9f3f53817e08a6a7f2264dd14de6a6a0c9db0df which is from my fork rather than your repository.

The following:
 https://github.com/biojppm/rapidyaml/blob/2cf09b073db1921938f7db87cb74560512829f55/samples/fetch_content/CMakeLists.txt#L17

should be updated to support cloning from the forked repository of a given PR.

biojppm:
Thanks for your work!

PR #393, opened 2023-11-12
fix: Checkout commit from forked repo

mboutet:
Possible fix for https://github.com/biojppm/rapidyaml/pull/392

### DIFF
--- a/samples/fetch_content/CMakeLists.txt
+++ b/samples/fetch_content/CMakeLists.txt
@@ -14,7 +14,7 @@ message(STATUS "FetchContent from branch: ${RYML_BRANCH_NAME}")
 
 include(FetchContent)
 FetchContent_Declare(ryml
-    GIT_REPOSITORY https://github.com/biojppm/rapidyaml.git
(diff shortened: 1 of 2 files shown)


samples/fetch_content/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ message(STATUS "FetchContent from branch: ${RYML_BRANCH_NAME}")
 
 include(FetchContent)
 FetchContent_Declare(ryml
-    GIT_REPOSITORY https://github.com/biojppm/rapidyaml.git
+    GIT_REPOSITORY https://github.com/${PULL_REQUEST_HEAD_REPO:biojppm/rapidyaml}.git
```
mboutet, line 17: I'm not super familiar with Make's syntax, so feel free to edit. I want to use `PULL_REQUEST_HEAD_REPO` and fallback to `biojppm/rapidyaml` if `PULL_REQUEST_HEAD_REPO` is not set.

biojppm: This should get you on your way: https://cmake.org/cmake/help/latest/variable/ENV.html

PR #407, opened 2024-03-06
Use _c4err macro for error message 'closing ] not found'

### LINKED ISSUE #406: Error message 'closing ] not found' does not use _c4err macro
All other calls to Parser::_err seem to come from the `_c4err` macro. I don't think that this is intentional considering it's the only direct invocation that I could find.

Running `grep -R -e '\(\s\|\.\|->\)_err(' rapidyaml` gives us all occurrances of `_err(` at a boundary:
```
rapidyaml/src/c4/yml/parse.cpp:            _err("closing ] not found");
rapidyaml/src/c4/yml/parse.hpp:    template<class ...Args> void _err(csubstr fmt, Args const& C4_RESTRICT ...args) const;
rapidyaml/src/c4/yml/detail/parser_dbg.hpp:         this->_err("ERROR:\n" "{}:{}: " fmt, __FILE__, __LINE__, ## __VA_ARGS__); } while(0)
rapidyaml/src/c4/yml/detail/parser_dbg.hpp:         this->_err("ERROR: " fmt, ## __VA_ARGS__); } while(0)
```

The first is this error message, the second is the definition of the `Parser::_err` method, and the third & fourth are the call within the `_c4err` macro.

WhiteAbeLincoln:
Fixes #406

### DIFF
--- a/src/c4/yml/parse.cpp
+++ b/src/c4/yml/parse.cpp
@@ -3569,7 +3569,7 @@ void Parser::_end_stream()
         _c4dbgpf("popping level: {} (stack sz={})", m_state->level, m_stack.size());
         _RYML_CB_ASSERT(m_stack.m_callbacks,  ! has_any(SSCL, &m_stack.top()));
         if(has_all(RSEQ|FLOW))
-            _err("closing ] not found");
+            _c4err("closing ] not found");
         _pop_level();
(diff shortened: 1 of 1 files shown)


biojppm:
Thanks!

PR #411, opened 2024-03-19
Improve error handling

### LINKED ISSUE #362: Use of debugbreak should be opt-in
There are two minor issues when including yml.hpp, when using sources from the 0.5.0 tag:

- I have to `#define C4_NO_DEBUG_BREAK` before inclusion otherwise in c4core (tag 0.11.0), a non-existing header c4/ext/debugbreak.h is included. 
- I have to `#pragma GCC diagnostic ignored "-Wold-style-cast"` to inhibit some g++ warnings.

While writing this ticket, I realized that the tar.gz of c4core 0.11.0 available on github does not contain files from external repositories, explaining why the header is missing. You may thus ignore this part (but why would a yaml library call `__debugbreak` ?).

For the second part, using `-Wno-old-style-cast` is of course specific to my project, but maybe others will use it and I have seen code to disable some warnings in rapidyaml, so maybe you will want to add this one.

Regards

### LINKED ISSUE #389: RoNodeMethods should have conditional noexcept
Throwing error callback
```
c4::yml::set_callbacks({ nullptr, nullptr, nullptr, [](const char*, std::size_t, c4::yml::Location, void*){ throw 42; } });
```

may lead to abort() in that case
```
ryml::ConstNodeRef node = Tree.rootref();
c4::csubstr invalidName = /*...*/;
node[invalidName];
```

bacuse noexcept [operator[]](https://github.com/biojppm/rapidyaml/blob/2cf09b073db1921938f7db87cb74560512829f55/src/c4/yml/node.hpp#L323) calls this callback 
```
    C4_ALWAYS_INLINE C4_PURE ConstImpl operator[] (csubstr k) const noexcept
    {
        _C4RV();
        size_t ch = tree_->find_child(id_, k);
        _RYML_CB_ASSERT(tree_->m_callbacks, ch != NONE);
        return {tree_, ch};
    }
```

I saw you mentioned that I can't return from error callback, so basically I can either call abort() or throw an exception, which in turn also leads to abort(). Config parsing error is an expected scenario in my case. So I'd like to handle such errors(even in debug build), rather than crashing the whole app

biojppm:
Fixes #389 
Fixes #362
Fix major error handling problem reported in [#389](https://github.com/biojppm/rapidyaml/issues/389) ([PR#411](https://github.com/biojppm/rapidyaml/pull/411)):

  - The `NodeRef` and `ConstNodeRef` classes had many methods marked `noexcept` that were doing assertions which could throw exceptions, causing an abort instead of a throw whenever the assertion called an exception-throwing error callback.
  - Also, this problem was compounded by exceptions being enabled in every build type -- despite the intention to have them only in debug builds. There was a problem in the preprocessor code to enable assertions which led to assertions being enabled in release builds even when `RYML_USE_ASSERT` was defined to 0. Thanks to @jdrouhard for reporting this.
  - Although the code is and was extensively tested, the testing was addressing mostly the happy path. Tests were added to ensure that the error behavior is as intended.
  - Together with this changeset, a major revision was carried out of the asserting/checking status of each function in the node classes. In most cases, assertions were added to functions cases that were missing them. So **beware** - user code that was invalid will now assert or error out. Also, assertions and checks are now directed as much as possible to the callbacks of the closest scope, ie if a tree has custom callbacks, errors should go through those callbacks.
  - Also, the intended assertion behavior is now in place: *no assertions in release builds*. **Beware** as well - user code which was relying on this will now silently succeed and return garbage in release builds. See the next points, which may help:
  - Added new methods to the `NodeRef`/`ConstNodeRef` classes:
    ```c++
    /** Distinguish between a valid seed vs a valid non-seed ref. */
    bool readable() const { return valid() && !is_seed(); }

    /** Get a child by name, with error checking; complexity is
     * O(num_children).
     *
     * Behaves as operator[](csubstr) const, but always raises an
     * error (even when RYML_USE_ASSERT is set to false) when the
     * returned node does not exist, or when this node is not
     * readable, or when it is not a map. This behaviour is similar to
     * std::vector::at(), but the error consists in calling the error
     * callback instead of directly raising an exception. */
    ConstNodeRef at(csubstr key) const;

    /** Get a child by position, with error checking; complexity is
     * O(pos).
     *
     * Behaves as operator[](size_t) const, but always raises an error
     * (even when RYML_USE_ASSERT is set to false) when the returned
     * node does not exist, or when this node is not readable, or when
     * it is not a container. This behaviour is similar to
     * std::vector::at(), but the error consists in calling the error
     * callback instead of directly raising an exception. */
    ConstNodeRef at(size_t pos) const;
    ```
  - Added macros and respective cmake options to control error handling:
    - `RYML_USE_ASSERT` - enable assertions regardless of build type. This is disabled by default.
    - `RYML_DEFAULT_CALLBACK_USES_EXCEPTIONS` - defines the same macro, which will make the default error handler provided by ryml throw exceptions instead of calling `std::abort()`. This is disabled by default.
  - Also, `RYML_DEBUG_BREAK()` is now enabled only if `RYML_DBG` is defined, as reported in [#362](https://github.com/biojppm/rapidyaml/issues/362).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,11 @@ c4_project(VERSION 0.5.0 STANDALONE
 
 option(RYML_WITH_TAB_TOKENS "Enable parsing of tabs after ':' and '-'. This is costly and disabled by default." OFF)
 option(RYML_DEFAULT_CALLBACKS "Enable ryml's default implementation of callbacks: allocate(), free(), error()" ON)
-option(RYML_BUILD_TOOLS "build tools" OFF)
+if(RYML_DEFAULT_CALLBACKS)
+    option(RYML_DEFAULT_CALLBACK_USES_EXCEPTIONS "Throw exceptions instead of calling abort in the default error handler provided by ryml" OFF)
+endif()
+option(RYML_USE_ASSERT "Enable assertions regardless of build type. Default is only when NDEBUG is not defined (which is in release builds). This causes a slowdown of the code." OFF)
+option(RYML_BUILD_TOOLS "Build tools" OFF)
 option(RYML_BUILD_API "Enable API generation (python, etc)" OFF)
 option(RYML_DBG "Enable (very verbose) ryml debug prints." OFF)
 
@@ -66,10 +70,18 @@ endif()
 
 if(NOT RYML_DEFAULT_CALLBACKS)
     target_compile_definitions(ryml PRIVATE RYML_NO_DEFAULT_CALLBACKS)
+else()
+    if(RYML_DEFAULT_CALLBACK_USES_EXCEPTIONS)
+        target_compile_definitions(ryml PRIVATE RYML_DEFAULT_CALLBACK_USES_EXCEPTIONS)
+    endif()
 endif()
 
 if(RYML_DBG)
     target_compile_definitions(ryml PRIVATE RYML_DBG)
+endif()
+
+if(RYML_USE_ASSERT)
+    target_compile_definitions(ryml PUBLIC RYML_USE_ASSERT=1)
 endif()
 
 
--- a/src/c4/yml/tree.cpp
+++ b/src/c4/yml/tree.cpp
@@ -246,10 +246,6 @@ ConstNodeRef Tree::rootref() const
     return ConstNodeRef(this, root_id());
 }
 
-ConstNodeRef Tree::crootref()
-{
-    return ConstNodeRef(this, root_id());
-}
 ConstNodeRef Tree::crootref() const
 {
     return ConstNodeRef(this, root_id());
@@ -257,23 +253,17 @@ ConstNodeRef Tree::crootref() const
 
 NodeRef Tree::ref(size_t id)
 {
-    _RYML_CB_ASSERT(m_callbacks, id != NONE && id >= 0 && id < m_size);
+    _RYML_CB_ASSERT(m_callbacks, id != NONE && id >= 0 && id < m_cap);
     return NodeRef(this, id);
 }
 ConstNodeRef Tree::ref(size_t id) const
 {
-    _RYML_CB_ASSERT(m_callbacks, id != NONE && id >= 0 && id < m_size);
-    return ConstNodeRef(this, id);
-}
-
-ConstNodeRef Tree::cref(size_t id)
-{
-    _RYML_CB_ASSERT(m_callbacks, id != NONE && id >= 0 && id < m_size);
+    _RYML_CB_ASSERT(m_callbacks, id != NONE && id >= 0 && id < m_cap);
     return ConstNodeRef(this, id);
 }
 ConstNodeRef Tree::cref(size_t id) const
 {
-    _RYML_CB_ASSERT(m_callbacks, id != NONE && id >= 0 && id < m_size);
+    _RYML_CB_ASSERT(m_callbacks, id != NONE && id >= 0 && id < m_cap);
     return ConstNodeRef(this, id);
 }
 
@@ -331,12 +321,12 @@ Tree::~Tree()
 }
 
 
-Tree::Tree(Tree const& that) noexcept : Tree(that.m_callbacks)
+Tree::Tree(Tree const& that) : Tree(that.m_callbacks)
 {
     _copy(that);
 }
 
-Tree& Tree::operator= (Tree const& that) noexcept
+Tree& Tree::operator= (Tree const& that)
 {
     _free();
     m_callbacks = that.m_callbacks;
@@ -344,12 +334,12 @@ Tree& Tree::operator= (Tree const& that) noexcept
     return *this;
 }
 
-Tree::Tree(Tree && that) noexcept : Tree(that.m_callbacks)
+Tree::Tree(Tree && that) : Tree(that.m_callbacks)
 {
     _move(that);
 }
 
-Tree& Tree::operator= (Tree && that) noexcept
+Tree& Tree::operator= (Tree && that)
 {
     _free();
     m_callbacks = that.m_callbacks;
@@ -1584,6 +1574,7 @@ size_t Tree::child(size_t node, size_t pos) const
 
 size_t Tree::child_pos(size_t node, size_t ch) const
 {
+    _RYML_CB_ASSERT(m_callbacks, node != NONE);
     size_t count = 0;
     for(size_t i = first_child(node); i != NONE; i = next_sibling(i))
     {
@@ -1821,7 +1812,7 @@ csubstr _transform_tag(Tree *t, csubstr tag, size_t node)
     size_t required_size = t->resolve_tag(substr{}, tag, node);
     if(!required_size)
         return tag;
-    const char *prev_arena = t->arena().str;
+    const char *prev_arena = t->arena().str; (void)prev_arena;
     substr buf = t->alloc_arena(required_size);
     _RYML_CB_ASSERT(t->m_callbacks, t->arena().str == prev_arena);
     size_t actual_size = t->resolve_tag(buf, tag, node);
--- a/README.md
+++ b/README.md
@@ -241,6 +241,514 @@ easily build and run this executable using any of the build samples,
 eg the [`add_subdirectory()` sample](samples/add_subdirectory/).
 
 ```c++
+// Parse YAML code in place, potentially mutating the buffer:
+char yml_buf[] = "{foo: 1, bar: [2, 3], john: doe}";
+ryml::Tree tree = ryml::parse_in_place(yml_buf);
+
+// The resulting tree contains only views to the parsed string. If
+// the string was parsed in place, then the string must outlive
+// the tree! This works in this case because `yml_buf` and `tree`
+// live on the same scope, so have the same lifetime.
+
+// It is also possible to:
+//
+//   - parse a read-only buffer using parse_in_arena(). This
+//     copies the YAML buffer to the tree's arena, and spares the
+//     headache of the string's lifetime.
+//
+//   - reuse an existing tree (advised)
+//
+//   - reuse an existing parser (advised)
+//
+// Note: it will always be significantly faster to parse in place
+// and reuse tree+parser.
+//
+// Below you will find samples that show how to achieve reuse; but
+// please note that for brevity and clarity, many of the examples
+// here are parsing in the arena, and not reusing tree or parser.
+
+
+//------------------------------------------------------------------
+// API overview
+
+// ryml has a two-level API:
+//
+// The lower level index API is based on the indices of nodes,
+// where the node's id is the node's position in the tree's data
+// array. This API is very efficient, but somewhat difficult to use:
+size_t root_id = tree.root_id();
+size_t bar_id = tree.find_child(root_id, "bar"); // need to get the index right
+CHECK(tree.is_map(root_id)); // all of the index methods are in the tree
+CHECK(tree.is_seq(bar_id));  // ... and receive the subject index
+
+// The node API is a lightweight abstraction sitting on top of the
+// index API, but offering a much more convenient interaction:
+ryml::ConstNodeRef root = tree.rootref();  // a const node reference
+ryml::ConstNodeRef bar = tree["bar"];
+CHECK(root.is_map());
+CHECK(bar.is_seq());
+
+// A node ref is a lightweight handle to the tree and associated id:
+CHECK(root.tree() == &tree); // a node ref points at its tree, WITHOUT refcount
+CHECK(root.id() == root_id); // a node ref's id is the index of the node
+CHECK(bar.id() == bar_id);   // a node ref's id is the index of the node
+
+// The node API translates very cleanly to the index API, so most
+// of the code examples below are using the node API.
+
+// WARNING. A node ref holds a raw pointer to the tree. Care must
+// be taken to ensure the lifetimes match, so that a node will
+// never access the tree after the goes out of scope.
+
+
+//------------------------------------------------------------------
+// To read the parsed tree
+
+// ConstNodeRef::operator[] does a lookup, is O(num_children[node]).
+CHECK(tree["foo"].is_keyval());
+CHECK(tree["foo"].val() == "1");   // get the val of a node (must be leaf node, otherwise it is a container and has no val)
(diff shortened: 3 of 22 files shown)


jdrouhard:
Tried your branch. Thanks for tackling this! One thing to note, you'll need to add some kind of void cast or `[[maybe_unused]]` attribute to some variables that are no longer used in release builds due to the fix of the assert macros. There are probably more, but at least this one triggered for me:

```gcc
src/c4/yml/parse.hpp:391:51: error: unused parameter 'str' [-Werror=unused-parameter]
    inline NodeData* _append_val_null(const char *str) { _RYML_CB_ASSERT(m_stack.m_callbacks, str >= m_buf.begin() && str <= m_buf.end()); return _append_val({nullptr, size_t(0)}); }
```

I had a small patch against an older commit that fixed all of them, but it probably doesn't merge quite cleanly any more:

```diff
diff --git a/src/c4/yml/parse.cpp b/src/c4/yml/parse.cpp
index a0f0dad..fbfc88c 100644
--- a/src/c4/yml/parse.cpp
+++ b/src/c4/yml/parse.cpp
@@ -4532,6 +4532,7 @@ bool Parser::_filter_nl(substr r, size_t *C4_RESTRICT i, size_t *C4_RESTRICT pos
 
     _RYML_CB_ASSERT(m_stack.m_callbacks, indentation != npos);
     _RYML_CB_ASSERT(m_stack.m_callbacks, curr == '\n');
+    (void) curr;
 
     _c4dbgfnl("found newline. sofar=[{}]~~~{}~~~", *pos, m_filter_arena.first(*pos));
     size_t ii = *i;
diff --git a/src/c4/yml/parse.hpp b/src/c4/yml/parse.hpp
index 659edf7..9a4993a 100644
--- a/src/c4/yml/parse.hpp
+++ b/src/c4/yml/parse.hpp
@@ -388,9 +388,9 @@ private:
     csubstr _consume_scalar();
     void  _move_scalar_from_top();
 
-    inline NodeData* _append_val_null(const char *str) { _RYML_CB_ASSERT(m_stack.m_callbacks, str >= m_buf.begin() && str <= m_buf.end()); return _append_val({nullptr, size_t(0)}); }
-    inline NodeData* _append_key_val_null(const char *str) { _RYML_CB_ASSERT(m_stack.m_callbacks, str >= m_buf.begin() && str <= m_buf.end()); return _append_key_val({nullptr, size_t(0)}); }
-    inline void      _store_scalar_null(const char *str) {  _RYML_CB_ASSERT(m_stack.m_callbacks, str >= m_buf.begin() && str <= m_buf.end()); _store_scalar({nullptr, size_t(0)}, false); }
+    inline NodeData* _append_val_null(const char *str) { _RYML_CB_ASSERT(m_stack.m_callbacks, str >= m_buf.begin() && str <= m_buf.end()); (void) str; return _append_val({nullptr, size_t(0)}); }
+    inline NodeData* _append_key_val_null(const char *str) { _RYML_CB_ASSERT(m_stack.m_callbacks, str >= m_buf.begin() && str <= m_buf.end()); (void) str; return _append_key_val({nullptr, size_t(0)}); }
+    inline void      _store_scalar_null(const char *str) {  _RYML_CB_ASSERT(m_stack.m_callbacks, str >= m_buf.begin() && str <= m_buf.end()); (void) str; _store_scalar({nullptr, size_t(0)}, false); }
 
     void  _set_indentation(size_t behind);
     void  _save_indentation(size_t behind=0);
diff --git a/src/c4/yml/tree.cpp b/src/c4/yml/tree.cpp
index b381986..fd8153d 100644
--- a/src/c4/yml/tree.cpp
+++ b/src/c4/yml/tree.cpp
@@ -1826,6 +1826,8 @@ csubstr _transform_tag(Tree *t, csubstr tag, size_t node)
     _RYML_CB_ASSERT(t->m_callbacks, t->arena().str == prev_arena);
     size_t actual_size = t->resolve_tag(buf, tag, node);
     _RYML_CB_ASSERT(t->m_callbacks, actual_size <= required_size);
+    (void) prev_arena;
+    (void) actual_size;
     return buf.first(actual_size);
 }
 void _resolve_tags(Tree *t, size_t node)
```

biojppm:
This is not complete yet. I intend to add tests to verify this, and I still need to search through the library for other places with `noexcept`. It will take a couple of weeks to get done.

Sorry for taking a bit long to look at this, I've been all hands on #414 which has been on the works for a long time; it is gigantic and really hard. So my attention is patchy here.

jdrouhard:
No worries. Just wanted to put it on your radar! Thanks again.

biojppm:
I sketched the assertion tests in test_tree.cpp. Feel free to continue if this is blocking you. Also, I think it is better to merge this before merging #414

biojppm:
@Neko-Box-Coder Thanks, that would be nice.

I did a major revision of the methods in the node classes. The following things are yet to be done:
  - Search the library for other `noexcept` places that could be hiding assertions/errors
  - Fix those methods and add tests similar to what was done so far.
  - Add tests to the `.at()` method recently added.
  - Add tests to cover the missing lines reported by Codecov.
  - Review the current changeset.

Pick your favorite!

biojppm:
> Search the library for other noexcept places that could be hiding assertions/errors

BTW, don't worry about the parser class. The parser has been rewritten in #414, and its contents went to different files, so changing it here (if required) would be pointless.

Neko-Box-Coder:
@biojppm 
I guess I overestimated my ability...kinda struggling to understand the codebase. No promises on how much I can help :sweat_smile: 

Anyway, I have replaced `noexcept` with `RYML_NOEXCEPT` in a few more places that are doing assert, not sure if it is correct or not.

Do I just push to your branch (not sure if I can do that) or do I create another PR to merge into this branch (`fix/389_noexcept`)?

biojppm:
> Do I just push to your branch (not sure if I can do that) or do I create another PR to merge into this branch (`fix/389_noexcept`)?

Thanks! I think it has to be a separate PR.

biojppm:
BTW the failing install tests are because of a dependency. Don't worry about those.

biojppm:
@Neko-Box-Coder @jdrouhard It's ready to be merged. Do you have any remarks?

CMakeLists.txt
```diff
@@ -66,12 +70,20 @@ endif()
...
+        target_compile_definitions(ryml PRIVATE RYML_DEFAULT_CALLBACK_USES_EXCEPTIONS)
+    endif()
 endif()
 
 if(RYML_DBG)
     target_compile_definitions(ryml PRIVATE RYML_DBG)
 endif()
 
+if(RYML_USE_ASSERT)
+    target_compile_definitions(ryml PUBLIC RYML_USE_ASSERT=1)
```
jdrouhard, line 84:
```suggestion
    target_compile_definitions(ryml PUBLIC RYML_USE_ASSERT)
```
should be sufficient.

jdrouhard:
I tried the current tip of this PR in our codebase and now I'm getting a ton of:

```
terminate called after throwing an instance of 'std::runtime_error'                                                                                                       
  what():  check failed: ((!(((Impl const* __restrict__)this)->is_seed())))
```

I haven't had a chance to dig in to see if I'm actually using the library incorrectly or not, but these same tests were passing previously. The actual changes here look good to me, though, so it's probably on my end. I'll confirm shortly.

CMakeLists.txt
biojppm: I did it again! Nice catch, thanks!

biojppm:
@jdrouhard remember that this PR not just fixes the `noexcept` problem, but it also adds thorough checking of the preconditions. Awaiting confirmation.

CMakeLists.txt
biojppm: Actually, no. I misread your suggestion and thought that you were proposing adding `=1` instead of removing it. 

Indeed, both are equivalent. But `=1` is more explicit, and consistent to what the value of the macro should be. I will stick to it.

jdrouhard (approved):

biojppm:
> Not sure it's totally clear that has_val() is not a valid call on a node that where valid() returns true, but changing the check to is_seed() works for me.

You are not the first to fall into this gotcha. In the PR I also improved the states of mutable refs, and now there is `invalid -> valid/seed -> valid/readable`:

```
invalid := not pointing at anything
valid: pointing at a tree/id pair, and further the node can be...
   readable := the node exists now                                     
   seed := the node may come to exist, if we write to it.
```

So both `readable` and `seed` are states where the node is also `valid`.

```c++
Tree tree = parse("{a: b}");
NodeRef invalid; // not pointing at anything.
NodeRef readable = tree["a"]; // also valid
NodeRef seed = tree["none"]; // also valid
```

I think the meaning of valid() became a bit vaguer  with this PR. Maybe even misleading. I am pondering deprecating it.

In your case, querying `has_val()` demands that the node is `.readable()` which is the same as (valid && !seed).

biojppm:
After thinking about it, I opted to keep .valid(), and improved the comments explaining this, and in the quickstart as well.

Neko-Box-Coder:
Hi @biojppm 
Thanks for the fix. I will review & test this out later today

biojppm:
> Not sure it's totally clear that has_val() is not a valid call on a node that where valid() returns true, but changing the check to is_seed() works for me.

I definitely agree that `.valid()` is misleading, as it can be construed to mean that you can use the `NodeRef` for anything, even reading, which is not the case.

I am not totally happy about this state of affairs.

So I just had an idea. I am thinking that `.valid()` could be deprecated, and there would be only `.invalid()`, and then `.is_seed()`, and `.readable()`. Because saying `! invalid()` is not as wide-ranging as saying `valid()`, so you become more aware that the other two states mean specific things. This may seem bike-shedding, but in my experience this `.valid()` prototype is exactly the sort of thing that spawns a thousand bug reports.

Picking `invalid()` instead of `valid()`. This way there would be less confusion, and less surprises. The reasoning about the node would be more direct.

Would appreciate feedback. @Neko-Box-Coder @jdrouhard what do you think?

Neko-Box-Coder:
@biojppm 
For me `valid()` or `invalid()` are both fine to me. But is there any reason to be called `is_seed()`? 
`is_seed()` is a bit abstract to me, would `readable()` and `writable()` work?

jdrouhard:
Works for me. Would the implicit cast to bool of a `NodeRef` return false in the seed state?

Assume `myNode["child"]` does not exist:

```cpp
if (auto child = myNode["child"]; !child) {
  child |= ryml::MAP;
  // populate a new map, etc....
}
```

The `!child` above was `!child.has_val()` prior to this PR, and now I have it as `child.is_seed()`. simply checking whether it's valid isn't enough here, because I don't want to repopulate the map if it already exists. Food for thought, I don't have a strong preference either way and I think what you described above is fine.

Neko-Box-Coder:
On second thought, yeah I think I get what you mean now regarding `valid()` vs `invalid()`.
It is quite a nuance difference I think but I do agree `invalid()` is less confusing

src/c4/yml/tree.cpp
```diff
@@ -331,25 +321,25 @@ Tree::~Tree()
...
     return *this;
 }
 
-Tree::Tree(Tree && that) noexcept : Tree(that.m_callbacks)
+Tree::Tree(Tree && that) : Tree(that.m_callbacks)
 {
     _move(that);
 }
 
-Tree& Tree::operator= (Tree && that) noexcept
```
Neko-Box-Coder, line 352: Would `RYML_NOEXCEPT` not work in here (and the rest of the constructors and operator=?

If `RYML_USE_ASSERT` is false, there will be no assert and therefore no exception, so the compiler can use choose move operator more easily.

Neko-Box-Coder (changes requested):

Neko-Box-Coder:
I have just finished reviewing it, only have 1 comment for the code.
I will test it now for my use case.

Also, would it be better to be more explicit for `valid()` / `invalid()`? 
For example, `valid_read()` or `valid_write()` (or `valid_seed()`) or `valid_read_write()`?

Neko-Box-Coder:
Yep, finished testing it. The exception catching is working and it seems good to me.

biojppm:
> Works for me. Would the implicit cast to bool of a `NodeRef` return false in the seed state?

Wow, I didn't remember that implicit cast. Thanks for raising the point.

That's another foot gun. I am going to deprecate that as well.

biojppm:
I started trying out this invalid/seed/readable idea, and it is really working so far. Already I caught a handful of cases where  `.valid()` was wrongly being used instead of `.readable()`.

It also helps that there is no semantic overlap between any of these three states.

Not done yet, but it's all for the better.

biojppm:
> Works for me. Would the implicit cast to bool of a `NodeRef` return false in the seed state?

@jdrouhard There is no implicit cast to bool. What I think is happening is that operator==(nullptr_t) is being invoked there. I have to investigate further, but I think that operator is too vague and ambiguous.

I am thinking of deprecating that too.

jdrouhard:
> > Works for me. Would the implicit cast to bool of a `NodeRef` return false in the seed state?
> 
> 
> 
> @jdrouhard There is no implicit cast to bool. What I think is happening is that operator==(nullptr_t) is being invoked there. I have to investigate further, but I think that operator is too vague and ambiguous.
> 
> 
> 
> I am thinking of deprecating that too.

To be fair, I was using the `!has_val()` version, I never actually tried to use a node's conversion to bool in a conditional like my example. I was just kinda guessing if that'd work 🙂

src/c4/yml/tree.cpp
biojppm: Good point. Indeed, ryml is only asserting inside these functions so RYML_NOEXCEPT could be used.

However if the user provides throwing allocation/free callbacks, then the noexcept would act as a barrier. So for now I think it's best to keep these without RYML_NOEXCEPT.

biojppm:
@jdrouhard indeed, there was no operator for implicit conversion to bool. That's something I would not do, as it is ambiguous, and with this API the point is to be specific. So that `if(node)` scenario fails to compile as it ought to.

I've finished with the `.invalid()` refactor, and I also deprecated `operator==(nullptr)` and `operator==(csubstr)`, as I said.

I just pushed it, so let's see whether the CI succeeds, but it is done if the CI succeeds.

biojppm:
Yep, green enough. Ready to merge.

PR #418, opened 2024-04-03
Replacing noexcept to RYML_NOEXCEPT for stack and tree

Neko-Box-Coder:
I have replaced `noexcept` for a few places with macro but not the ones that are move constructor.
All the ones I replaces have asserts which can throw (I think)

For #411


biojppm:
Thanks!

PR #419, opened 2024-04-03
Adding tests to at()

Neko-Box-Coder:
Trying to add tests to `ConstImpl at() const` in `node.hpp` for #411 

I have no idea if what I am doing is correct or not. And there's a `TODO` on line 1029 since I am not sure what to replace it with.
If this is totally wrong :sweat_smile: , I can close this PR.

I doubt I can help much with adding tests with my limited knowledge on the code base and test framework.
However, I can take a look at the whole PR and test it when it is finalized.

### DIFF
--- a/test/test_case.hpp
+++ b/test/test_case.hpp
@@ -175,6 +175,8 @@ struct ExpectError
     static void check_assertion(Tree *tree, std::function<void()> fn, Location expected={});
     static void check_success(std::function<void()> fn) { check_success(nullptr, fn); }
     static void check_success(Tree *tree, std::function<void()> fn);
+    static void verify_error(            std::function<void()> fn, Location expected={}) { verify_error(nullptr, fn, expected); }
+    static void verify_error(Tree *tree, std::function<void()> fn, Location expected={});
 };
 
 
--- a/test/test_case.hpp
+++ b/test/test_case.hpp
@@ -175,8 +175,6 @@ struct ExpectError
     static void check_assertion(Tree *tree, std::function<void()> fn, Location expected={});
     static void check_success(std::function<void()> fn) { check_success(nullptr, fn); }
     static void check_success(Tree *tree, std::function<void()> fn);
-    static void verify_error(            std::function<void()> fn, Location expected={}) { verify_error(nullptr, fn, expected); }
-    static void verify_error(Tree *tree, std::function<void()> fn, Location expected={});
 };
 
 
--- a/test/test_tree.cpp
+++ b/test/test_tree.cpp
@@ -848,6 +848,22 @@ void verify_assertion(csubstr src, Function &&fn)
     });
 }
 
+template<class Function>
+void verify_error(Tree &tree, Function &&fn)
+{
+    ExpectError::verify_error(&tree, [&]{
+        (void)fn(tree);
+    });
+}
+template<class Function>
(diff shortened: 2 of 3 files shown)


biojppm:
Nice, this saves me some work and helps! 

The only point is that verify_assertion() should be really the (non-existent) `verify_error()`, which in turn instead of `ExpectError::check_assertion()` would call `ExpectError::do_check()` with the same syntax. This is because we want to verify that those calls result in an error callback regardless of `RYML_USE_ASSERT`; `verify_assertion()` checks for the former.

Neko-Box-Coder:
Thanks for the feedback, I have added `verify_error()` to be used for `at()`. Makes sense that `at()` should always check (and callback) for any invalid inputs.

Do I need to add anything else?

test/test_case.hpp
```diff
@@ -175,6 +175,8 @@ struct ExpectError
     static void check_assertion(Tree *tree, std::function<void()> fn, Location expected={});
     static void check_success(std::function<void()> fn) { check_success(nullptr, fn); }
     static void check_success(Tree *tree, std::function<void()> fn);
+    static void verify_error(            std::function<void()> fn, Location expected={}) { verify_error(nullptr, fn, expected); }
```
biojppm, line 178: no need for this function here...

test/test_tree.cpp
```diff
@@ -848,6 +848,22 @@ void verify_assertion(csubstr src, Function &&fn)
     });
 }
 
+template<class Function>
+void verify_error(Tree &tree, Function &&fn)
+{
+    ExpectError::verify_error(&tree, [&]{
```
biojppm, line 854: ... just call directly `ExpectError::do_check()` here

test/test_tree.cpp
```diff
@@ -987,6 +1003,93 @@ TEST(Tree, operator_square_brackets)
...
+        EXPECT_FALSE(m.at(3)  != "3");
+        EXPECT_FALSE(m.at(4)  != "4");
+        EXPECT_FALSE(cm.at(0) != "0");
+        EXPECT_FALSE(cm.at(1) != "1");
+        EXPECT_FALSE(cm.at(2) != "2");
+        EXPECT_FALSE(cm.at(3) != "3");
+        EXPECT_FALSE(cm.at(4) != "4");
+        //
+        //TODO: Not sure what to replace with capacity
+        //verify_error(t, [&](Tree const&){ return cm[m.capacity()]; });
```
biojppm, line 1046: Asking for a node==capacity (which doesn't exist) should fail. Leave it here, I will fix it after merging.

biojppm (changes requested):

biojppm:
Thanks! Don't worry about the capacity. I will dig into it.

test/test_case.hpp
Neko-Box-Coder: Removed both `verify_error()`

test/test_tree.cpp
Neko-Box-Coder: Yep, updated it to call `do_check()` directly.

test/test_tree.cpp
Neko-Box-Coder: Cool

biojppm:
Thanks, it's good and it saved me some time!

PR #443, opened 2024-06-21
numbers: fix serialization with `+`, fix emit style heuristics

### LINKED ISSUE #442: serialization/deserialization behavior customization
Hello,

I just found out about this library today and I was just trying things out.  Overall I was able to get things working, however, there are some serialization/deserialization behaviors I would like to customize, but couldn't figure out how from the documentation.

For reference, my code is effectively doing the following:

```c++
// deserialization
auto content = c4::to_csubstr(data);  // convert from std::string_view to c4::csubstr
T obj;  // T is a scalar type like int, char, double, etc.
auto tree = ryml::parse_in_arena(content);
tree.rootref() >> obj;

// serialization
ryml::Tree out_tree;
out_tree.rootref() << obj;

std::ostream ostream(&buf);  // buf is a boost IO stream buffer that writes to a std::string
ostream << out_tree;
```

By trying various scalar values, what I noticed is that:
- If I pass in a negative number like `-123` or `-2.35e-10`, the serialized output string is always quoted with single quotes, like `'-123'` and `'-2.35e-10'`
- If I pass in a single space character with single quotes like `' '`, the output is a double quoted string like `" "`
- If I pass in a number with a plus sign like `+123` or `+3.5e+6`, it raises an exception.
- As far as I can tell from documentation, there is no overflow/underflow detection by default, but I couldn't figure out how.

I was wondering if I can tweak the parsing behavior, so that:
- for numbers, don't add any quotes.
- If it adds quotes around scalars, at least be consistent (either always use single quotes or double quotes).
- Ideally, support parsing the `+` sign in front of ints or doubles
- enable overflow/underflow detection.

Is it possible to customize the library with these changes, or do I need to use some adaptor/intermediate custom type tricks to do these customizations?  If I do need to use adaptors/custom types, is there some examples that I can look at?  Thanks!

biojppm:
fixes #442

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -1,3 +1,7 @@
 ## Fixes
 
+- Fix [#442](https://github.com/biojppm/rapidyaml/issues/442) ([PR#443](https://github.com/biojppm/rapidyaml/pull/443)):
+  - Ensure leading `+` is accepted when deserializing numbers.
+  - Ensure numbers are not quoted by fixing the heuristics in `scalar_style_query_plain()` and `scalar_style_choose()`.
+  - Add quickstart sample for overflow detection (only of integral types).
 - Parse engine: cleanup unused macros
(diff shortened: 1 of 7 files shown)


biojppm:
The failures in the becnchmarks are unrelated with this changeset. Merging now.

PR #444, opened 2024-06-22
fix: add /Zc:__cplusplus for correct __cplusplus value in MSVC

linuslau:
When compiling with C++17 in MSVC (VS2022), error C2039: 'chars_format': is not a member of 'std' occurs. This is due to the __cplusplus macro always being set to 199711L, regardless of the specified C++ standard. Consequently, conditionally included headers such as <charconv> are not included.

Adding the /Zc:__cplusplus option ensures the __cplusplus macro reflects the correct C++ standard. Please consider this change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,12 @@ if(CMAKE_COMPILER_IS_GNUCXX)
     endif()
 endif()
 
+if(MSVC)
+    option(ENABLE_ZC__cplusplus "Enable /Zc:__cplusplus compile option" ON)
+    if(ENABLE_ZC__cplusplus)
+        target_compile_options(ryml PUBLIC /Zc:__cplusplus)
+    endif()
+endif()
 
 #-------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


biojppm:
Thanks for the PR.

However, even if #445 is legitimate (let's discuss there), using a cmake option is not the proper way to go about it. Also, any fix must be in c4core, and not in this repo.

FYI, MSVC has the macro `_MSVC_LANG` which is defined to the proper value, and that is already handled in [`c4/language.hpp`](https://github.com/biojppm/c4core/blob/f9e970ac152034544d13dc3d9d109e17504274b1/src/c4/language.hpp#L20). At best, this macro would be used in the problematic place at `charconv.hpp`. Which, again, is a fix in c4core, not in this repo.

biojppm:
The fix to #445 was done in the c4core repo. See https://github.com/biojppm/c4core/commit/f4da68855bcb3015ec20c0c2f37180be90b7d268. 

Thanks again for the PR, but it does not address the underlying problem. Closing now.

PR #447, opened 2024-06-26
Plase use malloc.h instead of alloca.h on MinGW

toge:
Please refer c4core issue.
https://github.com/biojppm/c4core/pull/139


biojppm:
Thanks!

PR #452, opened 2024-07-15
fix: correct parser for double-quoted keys in map block

buty4649:
This PR fixes an issue where the single-quote parser was incorrectly specified for double-quoted keys in YAML map blocks.


biojppm:
Thanks! The CI failures are unrelated and need to be addressed separately. Merging now.

PR #454, opened 2024-07-21
Fix anchor reference issue in Map keys

buty4649:
This PR fixes a bug where an extra colon `:` is appended to the anchor reference name when used in Map keys.

### Reproduction Steps

Parsing the following `ng.yaml` with `./tools/ryml-yaml-events` results in an error.

```bash
$ cat ng.yaml
abc: &foo FOO
*foo: bar

$ ./build/tools/ryml-yaml-events ng.yaml
ng.yaml:2:7: (20B): ERROR: could not find ':' colon after key
ng.yaml:2:7: *foo: bar  (size=9)
                   ^~~  (cols 7-10)
```

### DIFF
--- a/src/c4/yml/parse_engine.def.hpp
+++ b/src/c4/yml/parse_engine.def.hpp
@@ -675,7 +675,7 @@ csubstr ParseEngine<EventHandler>::_scan_ref_map()
 {
     csubstr s = m_evt_handler->m_curr->line_contents.rem;
     _RYML_CB_ASSERT(m_evt_handler->m_stack.m_callbacks, s.begins_with('*'));
-    csubstr ref = s.first(s.first_of(",} "));
+    csubstr ref = s.first(s.first_of(",} :"));
     _line_progressed(ref.len);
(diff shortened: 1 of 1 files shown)


biojppm:
Sorry, but the existing behavior is correct, and adheres to the standard.

An anchor can legally have a trailing semicolon, so the semicolon immediately after a reference is understood as part of the reference. It is what it is.

See [example with space](https://play.yaml.io/main/parser?input=YWJjOiAmZm9vIEZPTwoqZm9vIDogYmFyCg==) and the same [example without space](https://play.yaml.io/main/parser?input=YWJjOiAmZm9vIEZPTwoqZm9vOiBiYXIK).

Also, if it weere o

buty4649:
Thank you for your reply and for explaining the details of the specification. I wasn't aware that an anchor could legally include a `:`. I had misunderstood because libyaml does not allow a `:` in this context.

As you pointed out, my changes are against the YAML specification, so I will close this PR.

PR #462, opened 2024-08-12
Fix CI

biojppm:


marcalff:
Possibly related:

* https://github.com/pypa/setuptools/issues/4478

Work around:

* https://github.com/pypa/setuptools/issues/4478#issuecomment-2235160778

PR #485, opened 2024-12-27
Generate GHA workflow files from originals

ingydotnet:
Moved original .github/workflows/*.yml files to a new dir called .github/workflows-ys/

Added a Makefile to the new directory. Running `make build` from there will generate the workflows/*.yml files from workflows-src/*.yml files using YAMLScript. Plain YAML files are valid YAMLScript, so these newly generated files should work the same.

The generated files have no comments and are formatted according to ys default rules.

From here forward, changes to the workflows-src files should not produce any changes to the workflow files.
This allows us to progressively refactor the workflow-src files and be sure no bad side effects have happened.


biojppm:
Thanks @ingydotnet ! This enables an effective way of working around github's limited workflow syntax, and makes maintenance of the workflows a breeze.

For example,
```
wc -lm --total=never gcc.ys gcc.yml
  185  6866 gcc.ys
  957 33146 gcc.yml
```
that is, 185 -> 957 lines, and 7KB -> 33KB. That's a huge saving!

PR #506, opened 2025-03-25
Fix merge key order, clear anchors after resolving

### LINKED ISSUE #400: Tree::resolve() leaves some references in place in certain cases
Consider the following document:
```
a: &a
  x: 1
b: &b
  ref: *a
c:
  ref: *b
```
Tree:resolve() will result in:
```
a:
  x: 1
b:
  ref:
    x: 1
c:
  ref:
    ref: &a
      x: 1
```
Which is correct safe for the unintended remaining '&a' reference. This happens whenever referenced nodes contains further references. As the reference instantiation process creates copies of the entire referenced sub-tree, including any references, there are now references that are not part of the list of anchors and references created at the start of resolve() and which is used to remove them at the end of resolve() - leaving out the 'newly created' references.

Workaround: call resolve() again (this will collect the additional 'newly created' references and delete them).

### LINKED ISSUE #484: Merge key order
Looks like the order is wrong for merge key

### input

```yaml
# $ cat /tmp/test5-simplified.yaml
base_1: &base_1
  a: 10
  b: 10
base_2: &base_2
  b: 20
k1:
  !!merge <<: *base_1
  !!merge <<: *base_2
```

### ryml

```
$ tools/ryml-parse-emit -j /tmp/test5-simplified.yaml  | jq .k1
reserving capacity=9
{
  "a": 10,
  "b": 10
}
```

### yq

```
$ yq 'explode(.)' /tmp/test5-simplified.yaml
base_1:
  a: 10
  b: 10
base_2:
  b: 20
k1:
  a: 10
  b: 20
```

biojppm:
Fixes #484 
Fixes #400

### DIFF
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -1,7 +1,8 @@
 
-## Fixes
+### Fixes
 
 - Fix [#400](https://github.com/biojppm/rapidyaml/issues/400) ([PR#506](https://github.com/biojppm/rapidyaml/pull/506)): clear anchors after resolving.
+- Fix [#484](https://github.com/biojppm/rapidyaml/issues/484) ([PR#506](https://github.com/biojppm/rapidyaml/pull/506)): fix merge key order for last element to be overriden.
 - [PR#503](https://github.com/biojppm/rapidyaml/pull/503): ensure parse error on `a: b: c` and similar cases containing nested maps opening on the same line.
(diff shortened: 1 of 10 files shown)


biojppm:
The coverage downgrade is likely spurious; merging now.

PR #515, opened 2025-04-01
Update python code in the readme

aarush-murari:
Python code only updated in test file afaik, but readme is not updated yet


biojppm:
Thanks for the fix, but #514 was already in place...

PR #542, opened 2025-09-15
Fix \x codepoint parsing, was not decoded

mutativesystems:
This part was unescaping `\xXX` incorrectly (it was just dumping the codepoint as a byte.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,7 +396,7 @@ ja: 惑星（ガス）
 zh: 行星（气体）
 # UTF8 decoding only happens in double-quoted strings,
 # as per the YAML standard
-decode this: "\u263A \xE2\x98\xBA"
+decode this: "\u263A c\x61f\xE9"
 and this as well: "\u2705 \U0001D11E"
 not decoded: '\u263A \xE2\x98\xBA'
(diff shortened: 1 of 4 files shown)


biojppm:
Thanks! Can you provide some examples where this would fail, and add some tests in [test/test_scalar_dquoted.cpp](https://github.com/biojppm/rapidyaml/tree/master/test/test_scalar_dquoted.cpp), say [here](https://github.com/biojppm/rapidyaml/blob/c90f8b361522f53c3d41e959a8e9f8a40b54d301/test/test_scalar_dquoted.cpp#L232) or [here](https://github.com/biojppm/rapidyaml/blob/c90f8b361522f53c3d41e959a8e9f8a40b54d301/test/test_scalar_dquoted.cpp#L408)?

biojppm (changes requested):
Can you fix the bad test in quickstart.cpp:912?

biojppm:
> Nevermind, the PR's CI isn't even checking out the correct commit...

Wow, that's astounding. It's weird as I'm using a very vanilla checkout action. I will investigate that elsewhere.

biojppm:
Thanks for sticking with this!

PR #553, opened 2025-10-24
allow build for MSVC clang-cl and some BSD systems

Sendan4:
Allow build ryml lib for MSVC Clang-cl compiler and some BSD systems (mainly freebsd). these systems and compilers dont have `<alloca.h> `header, but `alloca()` function is defined in `<stdlib.h>`.


Sendan4:
I'll see if I can modify my changes to make them look a little more consistent with this project.

PR #574, opened 2026-03-02
JSON emitting: improve handling of nan, inf and some floats

### LINKED ISSUE #312: Add dquotes around `inf` and `nan` when serializing to json
I think there is no way (as a number) to represent `Infinity` or `NaN` in JSON. One way would be to add them as strings.

Now what to do when converting to YAML... Should we keep the dquotes...

### LINKED ISSUE #535: Edge cases producing invalid JSON when emitting a tree read from YAML
I found some edge cases where reading YAML and emitting the resulting tree as JSON gives an invalid JSON document.

Take this test case:

```cpp
ryml::Tree tree;
ryml::NodeRef root = tree.rootref();
ryml::parse_in_arena(R"(a: inf
b: .inf
c: .1
d: 1.)",
    root);

std::cout << ryml::as_json(root);
std::cout << std::flush;
```

which gives

```
{"a": inf,"b": ".inf","c": .1,"d": 1.}
```

but if I read the YAML and JSON specs correctly, it would need to be

```
{"a": "inf","b": <special_string_for_inf>,"c": 0.1,"d": 1.0}
```

Reasons:

- Only `.inf` is recognized as the special infinity value.
- The `<special_string_for_inf>` should be whatever comes out of #312.
- JSON floats need to have digits on both sides of the decimal.

It seems to me that one would need to sanitize the data before emitting to JSON. AFAICT, you are already doing this for strings that need to be quoted in JSON.

I already reported problems with `inf` and `nan` here: https://github.com/biojppm/rapidyaml/issues/312#issuecomment-3048092744. Let me know if I can be of help. I had a look through the test suite and known limitations and could not find anything related to these problems. If you agree this should be fixed, I can give it a try and open a PR :) 

FYI @davidrudlstorfer @gilrrei

biojppm:
For example, the tree
  ```yaml
  {
    inf: [inf, .inf, .Inf, .INF, -inf, -.inf, -.Inf, -.INF],
    nan: [nan, .nan, .NaN, .NAN],
    dot: [.1, 1., .2e2, 10., -.2, -2.],
    zero: [10, 01],
    normal: [0.1, 0.2e3, 4.e5],
  }
  ```
is now emitted to JSON as:
  ```json
  {
    "inf": [".inf",".inf",".inf",".inf","-.inf","-.inf","-.inf","-.inf"],
    "nan": [".nan",".nan",".nan",".nan"],
    "dot": [0.1,1.0,0.2e2,10.0,-0.2,-2.0],
    "zero": [10,"01"],
    "normal": [0.1,0.2e3,4.e5]
  }
  ```
Previously, some inf and nan cases were emitted without quotes. Note also the added zeroes for some floats, eg `.1` or `-2.`, which turn into `0.1` and `-2.0` in JSON.

I tried using "8e888" for inf as suggested in #312 but this value cannot be parsed by to_chars() (see proof in added tests), which prevents roundtrip equivalence. So I opted for a sensible `".inf"` string for all inf cases.

Fixes #312 
Fixes #535


biojppm:
@davidrudlstorfer @captain-yoshi @sebproell Tagging you here because of #535 and #312 .

Please take a look at this PR; IMO it addresses the problems discussed in those issues.

As I said above (and you can verify in the tests added in this PR), `8e888` cannot be used to signify infinity because fast_float `to_chars()` cannot parse this, which prevents roundtrip equivalence. I did not look further than that into this aspect.

Other than that, the proposed JSON strings `.nan`, `.inf` and `-.inf` can be used say in JS to check for these special values, and ryml deserializes them to their proper values as well.

PR #585, opened 2026-03-29
Ensure parse errors on more invalid YAML cases

biojppm:


biojppm:
The line not covered was from before and not introduced in this PR. Merging now.

PR #599, opened 2026-05-08
Include `c4Project.cmake` after calling `project()`

### LINKED ISSUE #256: Installation directory on Linux 64bit OS
I found that when run `make install`, the binary files are going to `/usr/lib` instead of `/usr/lib64`, which is the default lib directory on 64bit Linux OS.

Can it be changed to `/usr/lib64`? Thanks!

musicinmybrain:
This fixes the `GNUInstallDirs` module not working as expected; see discussion in https://github.com/biojppm/cmake/pull/16.

Fixes #256

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,11 @@
 cmake_minimum_required(VERSION 3.12)
-include(./ext/c4core/cmake/c4Project.cmake)
 cmake_policy(SET CMP0048 NEW) # pass VERSION to project()
 project(ryml
     VERSION 0.12.1
     DESCRIPTION "Rapid YAML parsing and emitting"
     HOMEPAGE_URL "https://github.com/biojppm/rapidyaml"
     LANGUAGES CXX)
+include(./ext/c4core/cmake/c4Project.cmake)
 include(./compat.cmake)
 
 c4_project(VERSION 0.12.1 STANDALONE
--- a/changelog/current.md
+++ b/changelog/current.md
@@ -1,0 +1,8 @@
+- Fix [256](https://github.com/biojppm/rapidyaml/issues/256): installation directory on Linux 64bit ([PR#599](https://github.com/biojppm/rapidyaml/pull/599)). See also original fix at [cmake#16](https://github.com/biojppm/cmake/pull/16). Big thanks to @GabrielBarrantes and @musicinmybrain, not just for their fixes but also for all their downstream work!
+
+
+### Thanks
+
+ - @GabrielBarrantes
+ - @musicinmybrain
+
--- a/samples/custom_c4core/CMakeLists.txt
+++ b/samples/custom_c4core/CMakeLists.txt
@@ -1,10 +1,10 @@
 cmake_minimum_required(VERSION 3.12)
 
-include(../../ext/c4core/cmake/c4Project.cmake)
 project(ryml-quickstart
     DESCRIPTION "Shows how to use a custom c4core version,
                  different from ryml's c4core submodule"
     LANGUAGES CXX)
+include(../../ext/c4core/cmake/c4Project.cmake)
 c4_project(VERSION 0.1.0 STANDALONE
     AUTHOR "Joao Paulo Magalhaes <dev@jpmag.me>")
 
--- a/samples/quickstart.cpp
+++ b/samples/quickstart.cpp
@@ -2217,21 +2217,26 @@ cars: GTO
             keys.emplace_back(n.key());
             vals.emplace_back(n.has_val() ? n.val() : ryml::csubstr{});
         }
-        CHECK(keys[0] == "doe");
-        CHECK(vals[0] == "a deer, a female deer");
-        CHECK(keys[1] == "ray");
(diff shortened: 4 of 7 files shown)


biojppm:
The coverage CI failures are unrelated, they are due to CI misconfiguration (trying to upload coverage without token on the PR-submitting repo; this needs to be fixed elsewhere). 

The gcc13 canary failures are meaningful, however. [Looking at the failure](https://github.com/biojppm/rapidyaml/actions/runs/25548279788/job/74989708254?pr=599#step:7:132), and seeing the `-Werror=null-dereference` compilation errors, it means that the placement of `include(c4Project)` WRT `project()` is playing a role in the compilation flags!

I also saw this same behavior in a different branch, after moving the include to after project(). (EDIT: in the armv5 compilation, but interestingly not in the same source file, although it is equal in both branches).

I will push a number of fixes to this PR to turn these jobs to green.

biojppm:
Here are some relevant findings.

With g++-13, moving `include(c4Project)` to after `project()` enables the following additional warnings:

`-Wunused-const-variable -Wignored-attributes -Wnull-dereference -Wmisleading-indentation -Wduplicated-cond -Wduplicated-branches -Waddress-of-packed-member`.

[Tracing this in `c4Project.cmake`](https://github.com/biojppm/cmake/blob/10581a42ce4667b407dde5858dfd0be95afb00b6/c4Project.cmake#L917) there's this interesting block:

```cmake
if(CMAKE_CXX_COMPILER_ID STREQUAL "")
    message(FATAL_ERROR "project() must be called before including this file")
elseif(CMAKE_CXX_COMPILER_ID STREQUAL GNU)
    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 6.0)
        list(APPEND _C4_PEDANTIC_FLAGS_GCC
            -Wunused-const-variable
            -Wignored-attributes
            -Wnull-dereference # warn if a null dereference is detected
            -Wmisleading-indentation # where indentation implies blocks where blocks do not exist
            -Wduplicated-cond # where if-else chain has duplicated conditions
            $<$<COMPILE_LANGUAGE:C>:-Wabsolute-value>
        )
    endif()
    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7.0)
```

Let's start by addressing the second branch. It was not being hit, meaning at least one of `CMAKE_CXX_COMPILER_ID` or `CMAKE_CXX_COMPILER_VERSION` was not adequately set. Moving `include(c4Project)` fixes this and now the branch is hit. Now we get the compiler flags as intended. Nice!

---

But the first branch is more interesting. First off, it tells us that moving `include(c4Project)` is absolutely the right thing. Next, **the fatal error should have been firing but wasn't**.

Why?

To investigate, I moved back `include(c4Project)` to its original wrong place and added this print:
```diff
modified   c4Project.cmake
@@ -914,6 +914,8 @@ set(_C4_PEDANTIC_FLAGS_GCC ${_C4_PEDANTIC_FLAGS_COMMON}
     -Wuseless-cast # cast to the same type
     )
 
+message(STATUS "compiler: '${CMAKE_CXX_COMPILER_ID}' - '${CMAKE_CXX_COMPILER_VERSION}'")
 if(CMAKE_CXX_COMPILER_ID STREQUAL "")
     message(FATAL_ERROR "project() must be called before including this file")
 elseif(CMAKE_CXX_COMPILER_ID STREQUAL GNU)
```
... resulting in this:
```
-- compiler: '' - ''
-- The CXX compiler identification is GNU 13.3.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
```

So `CMAKE_CXX_COMPILER_ID` prints as `''`, but `if(CMAKE_CXX_COMPILER_ID STREQUAL "")` is false!!! 

That is, the check doesn't work. It's an unfortunate quirk of cmake. I changed it thus:
```diff
modified   c4Project.cmake
@@ -914,7 +914,7 @@ set(_C4_PEDANTIC_FLAGS_GCC ${_C4_PEDANTIC_FLAGS_COMMON}
     -Wuseless-cast # cast to the same type
     )
 
+message(STATUS "compiler: '${CMAKE_CXX_COMPILER_ID}' - '${CMAKE_CXX_COMPILER_VERSION}'")
-if(CMAKE_CXX_COMPILER_ID STREQUAL "")
+if((NOT CMAKE_CXX_COMPILER_ID) OR (CMAKE_CXX_COMPILER_ID STREQUAL ""))
     message(FATAL_ERROR "project() must be called before including this file")
 elseif(CMAKE_CXX_COMPILER_ID STREQUAL GNU)
     if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 6.0)
```
and it now works as intended, causing an error: 
```
-- compiler: '' - ''
CMake Error at ext/c4core/cmake/c4Project.cmake:919 (message):
  project() must be called before including this file
  CMakeLists.txt:3 (include)
```
So that change needs to be added to `c4Project` too.

biojppm:
Thanks @musicinmybrain and @GabrielBarrantes for your help fixing this! :100:

PR #610, opened 2026-05-26
Fix some clang++ warnings

TedLyngmo:
Compiled with `-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic` and fixed some low hanging fruits.

### DIFF
--- a/src/c4/yml/event_handler_tree.hpp
+++ b/src/c4/yml/event_handler_tree.hpp
@@ -247,11 +247,11 @@ public:
     /** @name YAML map events */
     /** @{ */
 
-    void begin_map_key_flow()
+    C4_NORETURN void begin_map_key_flow()
     {
         _RYML_ERR_PARSE_(m_stack.m_callbacks, m_curr->pos, "ryml trees cannot handle containers as keys");
     }
-    void begin_map_key_block()
+    C4_NORETURN void begin_map_key_block()
     {
         _RYML_ERR_PARSE_(m_stack.m_callbacks, m_curr->pos, "ryml trees cannot handle containers as keys");
     }
@@ -297,11 +297,11 @@ public:
     /** @name YAML seq events */
     /** @{ */
 
-    void begin_seq_key_flow()
+    C4_NORETURN void begin_seq_key_flow()
     {
         _RYML_ERR_PARSE_(m_stack.m_callbacks, m_curr->pos, "ryml trees cannot handle containers as keys");
     }
-    void begin_seq_key_block()
+    C4_NORETURN void begin_seq_key_block()
     {
         _RYML_ERR_PARSE_(m_stack.m_callbacks, m_curr->pos, "ryml trees cannot handle containers as keys");
     }
@@ -386,7 +386,7 @@ public:
      * See the documentation for @ref doc_event_handlers, which has
      * important notes about this event.
      */
-    void actually_val_is_first_key_of_new_map_block()
+    C4_NORETURN void actually_val_is_first_key_of_new_map_block()
     {
         _RYML_ERR_PARSE_(m_stack.m_callbacks, m_curr->pos, "ryml trees cannot handle containers as keys");
     }
--- a/src/c4/yml/node.hpp
+++ b/src/c4/yml/node.hpp
@@ -725,7 +725,7 @@ public:
     {
         _C4RR();
         return from_chars(val(), &v);
-    };
+    }
 
     /** @} */
 
--- a/src/c4/yml/parse_engine.def.hpp
+++ b/src/c4/yml/parse_engine.def.hpp
@@ -1134,6 +1134,7 @@ bool ParseEngine<EventHandler>::_scan_scalar_plain_seq_flow(ScannedScalar *C4_RE
                     _c4err("parse error"); // no return
                 }
             }
+            break;
         default:
             ;
         }
--- a/src/c4/yml/parse_engine.def.hpp
+++ b/src/c4/yml/parse_engine.def.hpp
@@ -463,7 +463,8 @@ void ParseEngine<EventHandler>::_relocate_arena(csubstr prev_arena, substr next_
     if((s).str >= pb && (s).str <= pe)              \
     {                                               \
         (s).str = next_arena.str + ((s).str - pb);  \
-    }
+    }                                               \
+    ((void)0)
     for(ParserState &st : m_evt_handler->m_stack)
     {
         _ryml_relocate(st.line_contents.rem);
@@ -3584,11 +3585,7 @@ void ParseEngine<EventHandler>::_filter_block_folded_indented_block(FilterProces
                 {
                     const char c = rem[first];
                     _c4dbgfbf("firstns={}='{}'", first, _c4prc(c));
-                    if(c == '\n' || c == '\r')
-                    {
-                        ;
-                    }
-                    else
+                    if(c != '\n' && c != '\r')
                     {
                         _c4dbgfbf("done with indented block",  first);
                         goto endloop;
--- a/src/c4/yml/parse_engine.def.hpp
+++ b/src/c4/yml/parse_engine.def.hpp
@@ -1616,7 +1616,7 @@ substr ParseEngine<EventHandler>::_peek_next_line(size_t pos) const
 
 next_is_empty:
     _c4dbgpf("peek next line @ {}: (len=0)''", pos);
-    return {};
+    return rem;
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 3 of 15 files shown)


src/c4/yml/node.hpp
```diff
@@ -1022,11 +1023,13 @@ class RYML_EXPORT NodeRef : public detail::RoNodeMethods<NodeRef, ConstNodeRef>
     // require valid: a helper macro, undefined at the end
     #define _C4RR()                                                     \
         _RYML_ASSERT_BASIC(m_tree != nullptr);                           \
-        _RYML_ASSERT_VISIT_(m_tree->m_callbacks, m_id != NONE && !is_seed(), m_tree, m_id)
+        _RYML_ASSERT_VISIT_(m_tree->m_callbacks, m_id != NONE && !is_seed(), m_tree, m_id); \
+        ((void)0)
```
biojppm: Why not remove the semicolon at the end of `_RYML_ASSERT_VISIT_()`? Wouldn't that spare the need for adding `((void)0)`?

src/c4/yml/node.hpp
```diff
@@ -1022,11 +1023,13 @@ class RYML_EXPORT NodeRef : public detail::RoNodeMethods<NodeRef, ConstNodeRef>
...
     // require id: a helper macro, undefined at the end
     #define _C4RID()                                                        \
         _RYML_ASSERT_BASIC(m_tree != nullptr);                                     \
-        _RYML_ASSERT_VISIT_(m_tree->m_callbacks, m_id != NONE, m_tree, m_id)
+        _RYML_ASSERT_VISIT_(m_tree->m_callbacks, m_id != NONE, m_tree, m_id); \
+        ((void)0)
```
biojppm: Same as above

src/c4/yml/node.hpp
biojppm: I see now that both the semicolon and the void no-op were added. What's the reasoning behind this?

src/c4/yml/event_handler_tree.hpp
```diff
@@ -247,11 +247,11 @@ struct EventHandlerTree : public EventHandlerStack<EventHandlerTree, EventHandle
     /** @name YAML map events */
     /** @{ */
 
-    void begin_map_key_flow()
+    C4_NORETURN void begin_map_key_flow()
```
biojppm, line 250: Did the warnings require this? Not contesting the change itself as it is sane.

biojppm (commented):
Thanks for the PR, it is really appreciated. 

First, some context: [-Weverything](https://stackoverflow.com/a/64147755) is not actually advised by clang.

There are a couple of changeset remarks/questions above. Other than these, there are a few others:
 - the initial diff had a rather large part fixing leading underscores in some symbols; subsequently they're not here. Why the revert?
 - Also, on this same subject, I think suppressing the warning as needed with file-wide push/pop is not just lower effort, but might actually be preferable. IMO there's really low risk of having these symbols clash with any other symbol.
- what are the final compile flags that you're looking to suppress? I think adding these to the CI is in order, but I'm a bit confused as to what the flags are based on the PR description and commit error messages.
- related - were there no warnings coming from c4core code? I would really expect that to be the case, but there's no PR there.

TedLyngmo:
> First, some context: [-Weverything](https://stackoverflow.com/a/64147755) is not actually advised by clang. But I'm really inclined

Yes, using it means you have no control over the warnings that will be enabled. Every new version of clang++ will come with new warnings. I use it as a means to enable everything in the current version of clang++ that I have - and then remove it once I've fixed everything that popped up and looked interesting enough :) Shipping a library with `-Weverything` enabled is not a good idea.

> There are a couple of changeset remarks/questions above. Other than these, there are a few others:
> 
> * the initial diff had a rather large part fixing leading underscores in some symbols; subsequently they're not here. Why the revert?

I misunderstood the large impact and concentrated on just being able to compile the library clean. As it happens, CI showed that there is a lot more code that depends on those identifiers :-) I removed that commit from the set and was thinking of addressing that separately now that I have a better picture of how they are used.

> * Also, on this same subject, I think suppressing the warning as needed with file-wide push/pop is not just lower effort, but might actually be preferable. IMO there's really low risk of having these symbols clash with any other symbol.

True, but there are organizations that are sticklers when it comes to undefined behavior and they'd probably reject the library when noticing these identifiers. I have seen header guards and other reserved identifiers defined in libraries collide with the implementation's own identifiers after upgrading the toolchain. Not a nice surprise.

> * what are the final compile flags that you're looking to suppress? I think adding these to the CI is in order, but I'm a bit confused as to what the flags are based on the PR description and commit error messages.

I wasn't thinking of suppressing any. Possibly adding an explicit set of warnings after having fixed those specific issues.

> * related - were there no warnings coming from c4core code? I would really expect that to be the case, but there's no PR there.

Oh, there were tons. Since it's in a submodule, I left those untouched. It'd probably be a good idea to go through those separately.

src/c4/yml/node.hpp
TedLyngmo: I can only explain that with "it was late last night" :-)

src/c4/yml/event_handler_tree.hpp
TedLyngmo: Yes, the lack of it triggered `-Wmissing-noreturn`.

biojppm:
> >   what are the final compile flags that you're looking to suppress? I think adding these to the CI is in order, but I'm a bit confused as to what the flags are based on the PR description and commit error messages.
> 
> I wasn't thinking of suppressing any. Possibly adding an explicit set of warnings after having fixed those specific issues.

My bad -- I meant *fix* instead of *suppress*. The question still stands. What are those warnings? The PR description says one thing but individual commits show another.

src/c4/yml/node.hpp
biojppm: Please revert (based on comment below)

src/c4/yml/node.hpp
biojppm: Ok. Please revert this then.

src/c4/yml/parse_engine.def.hpp
```diff
@@ -35,7 +35,8 @@
     this->_err(RYML_LOC_HERE(), __VA_ARGS__)
 #endif
 #define _c4assert(...)   \
-    _RYML_ASSERT_PARSE_(m_evt_handler->m_stack.m_callbacks, __VA_ARGS__, m_evt_handler->m_curr->pos)
+    _RYML_ASSERT_PARSE_(m_evt_handler->m_stack.m_callbacks, __VA_ARGS__, m_evt_handler->m_curr->pos); \
+    ((void)0)
```
biojppm: Please revert.

TedLyngmo:
> > > what are the final compile flags that you're looking to suppress? I think adding these to the CI is in order, but I'm a bit confused as to what the flags are based on the PR description and commit error messages.
> > 
> > I wasn't thinking of suppressing any. Possibly adding an explicit set of warnings after having fixed those specific issues.
> 
> My bad -- I meant _fix_ instead of _suppress_. The question still stands. What are those warnings? The PR description says one thing but individual commits show another.

Each commit is supposed to only fix the warning(s) mentioned in the four individual commit messages. For the PR, it could say _"Fix clang++ warnings: -Wmissing-noreturn, -Wextra-semi, -Wextra-semi-stmt, -Wimplicit-fallthrough and -Wnrvo"_ but that'd require updating the PR message if I pull anything out of the set (or add anything to it).

biojppm:
Looking at the [failing windows tests](https://github.com/biojppm/rapidyaml/actions/runs/26492676288/job/78078892474?pr=610#step:5:4933) : the newly-added  `C4_NORETURN` to the tree handler events now causes a `warning C4702: unreachable code` on MSVC.

This is triggered from an engine test that checks for an error (ie requires an error) in the specified sequence of event handler calls when using the tree handler. That test must remain.

IMO the way to tackle this is to work around it by suppressing that warning it in the offending test source code: adding `C4_SUPPRESS_WARNING_MSVC_WITH_PUSH(4702)` and corresponding `_POP` at the top/bottom of the files. 

AFAICT these are the files where that would be needed:

  - test/test_engine_3_map_block_2.cpp
  - test/test_engine_3_map_flow_1.cpp
  - test/test_engine_4_anchor.cpp
  - test/test_engine_5_tag_1.cpp
  - test/test_engine_6_qmrk_1.cpp
  - test/test_engine_6_qmrk_2.cpp
  - test/test_engine_6_qmrk_3.cpp
  - test/test_engine_6_qmrk_4.cpp
  - test/test_engine_7_seqimap.cpp

HTH.

biojppm:
The failure in the coverage workflows is unrelated with this PR, and can be ignored. It's a token access problem that must be fixed elsewhere. 

The install singleheader failure is a flaky test ; also unrelated.

biojppm (approved):

biojppm:
I pushed a bunch of branches before approving this run. It's going to take a while.

biojppm:
Thanks!

PR #616, opened 2026-06-07
API cleanup: writers and error_on_excess

biojppm:
- `WriterFile` and `WriterOStream` no longer track the number of emitted bytes.
  - `error_on_excess` is now used in the emit-to-buffer overloads, and no longer in the main `Emitter::emit_as()` driver function.


biojppm:
The coverage miss was not introduced in this PR. Merging now.

PR #512, opened 2025-03-30
WIP no submodules

biojppm:
This PR removes c4core as a submodule. It is split into two sub-requests
  - Part 1 #630:  This only copies the c4core source code and required project files. Just copy, no logic changes.
  - Part 2 #631: adjust this project to pass the CI. Focus on this one.

---
From the [comments below](https://github.com/biojppm/rapidyaml/pull/512#issuecomment-3865697023):

> If I understand correctly, the c4core submodule is replaced by a copy of the c4core repository, correct ?

Yes, with some details to it:

  - this PR removes the c4core submodule, and copies c4core code into the rapidyaml tree, to two different dirs:
     - `ext/c4core.src`: c4core code used in the rapidyaml library 
     - `ext/c4core.dev`: c4core code used by tests/benchmarks
    Doing this makes the standalone library marginally smaller, and enforces the standalone default of rapidyaml bypassing the submodule. So most users (which don't care about c4core itself) will now see rapidyaml as a plain project without extra dependencies or submodules.
  - this PR retains the existing option to compile against a full instance of the c4core library, thus ignoring the in-source copy under `ext/c4core.src` and `ext/c4core.dev`
    - the default is still to compile in standalone mode (ie c4core part of rapidyaml)
    - the pre-existing cmake option `RYML_STANDALONE` still enables users to build rapidyaml against a previously build c4core library (like in the ubuntu packages where rapidyaml depends on c4core).
  - the amalgamation tool should work just as before

---

> Why is this desirable, and what issue does this resolve ?

I have had repeated feedback that the c4core submodule makes it impossible for some people to use rapidyaml, either because it contains submodules, or because c4core is a dependency boogeyman. 

While I disagree with both of these reasonings, it is not for me to dispute their choice.

So I am removing these reasons by making the c4core code really standalone in rapidyaml, while striving to retain all current existing workflows.

---
> 
> Do you plan to remove the c4core repository and only maintain it within rapidyaml

**No, not at all. c4core will remain just as it is, in https://github.com/biojppm/c4core.**

> and if not, how will the fixes from c4core be applied to rapidyaml (beside manually) ?

This PR adds a helper makefile to simplify the workflow, with two main approaches:

 - create an in-source clone of c4core (parallel to the existing copies, under the old location of the submodule `ext/c4core`) and...
 - importing changes from c4core clone to rapidyaml copies
 - exporting changes from rapidyaml copies to c4core clone
 - check sync status.

[...]

Issue #123, opened 2021-03-09
Feature request: flat repository structure


kerrick-lyft:
Hi! :wave:

I was recently evaluating ryml vs [yaml-cpp](https://github.com/jbeder/yaml-cpp) for a project and I decided to go with yaml-cpp, and I thought maybe the reasoning behind this decision would be useful feedback.

Our use-case is that we use YAML to parse data files for tests in CI. These data files can be up to ~500kb each, and we will have 10 or 20 of them (TBD). ryml could save us maybe 5-10 seconds per CI run, which, all else being equal, would be good.

However, because of ryml's code structure (recursively nested Git submodules), the time to pull the code dominates any runtime benefits. Our build system ([Bazel](https://bazel.build/)) does roughly the following in order to clone the ryml code:
```
git clone https://github.com/biojppm/rapidyaml.git --shallow-since='1613428986 +0000' \
&& cd rapidyaml \
&& git submodule update --init --recursive
```
On my machine, this takes about 18 seconds to run.

To fetch yaml-cpp, this is what Bazel does:
```
curl -L https://github.com/jbeder/yaml-cpp/archive/a6bbe0e50ac4074f0b9b44188c28cf00caf1a723.tar.gz \
| tar -xz
```
On my machine, this takes about 2 seconds to run. Note that we can skip dealing with git because all the code is in a single repo, and instead just fetch the tarball from GitHub.

Someday in the future we might configure caching for build inputs, but even then I think Bazel only supports caching binary blobs (like yaml-cpp's `.tar.gz` file) and not Git repositories, so this cost won't go away (it will in fact get worse compared to yaml-cpp).

To get parity with yaml-cpp here, the rapidyaml repo could have a  copy of the c4 source instead of using a submodule. Obviously this has downsides in terms of how easy it is for you to maintain the code in this repo.

Just a suggestion, hopefully you think it's a useful idea :smile:

Gei0r:
Are you re-building your build tools on every CI run? That seems unreasonable.

kerrick-lyft:
We re-compile the source code for our unit tests, which depend on ryml.

biojppm:
Hi @kerrick-lyft, and thanks for the suggestion.

Indeed I really prefer the submodule approach, even if it is nested. I recognize that the clone time may be a downside, but OTOH any other meaningful alternative that continues to be easy for me to track changes would be too much work. As it is my time is rather limited, so I will have to skip your suggestion.

[...]

PR #636, opened 2026-06-23
Remove submodules

biojppm:
This PR removes the c4core submodule, and copies c4core code into the rapidyaml tree:
  - `ext/c4core.src/`: c4core code used in the rapidyaml library 
  - `ext/c4core.dev/`: c4core code used by tests/benchmarks
  - `proj/`: project files like cmake files and toolchains
  - `RYML_STANDALONE` can still be set to `OFF` to compile against an out-of-source c4core
  - the amalgamation tool now amalgamates only the files in `c4core.src` files, and has new option `--c4core-dev`  to include the `c4core.dev` files
  - adds tools to manage c4core synchronization scenarios. See [ext/README.md](https://github.com/biojppm/rapidyaml/tree/master/ext/) for details with this

--- 

Because copying the c4core files created a large volume of code, it is impossible to review this PR. So this PR is split into two sub-PRs:
  - Part 1 #630:  This only copies the c4core source code and required project files. Just copy, no logic changes at all. This PR cannot pass any CI checks.
  - Part 2 #631: changes required to pass the CI. **That is the PR that should be reviewed.** 

This PR was formerly at #512 (see relevant comments there), but that one was closed by github when the master branch got in sync with it. So I had to open this PR.

---

## Why is this desirable, and what issue does this resolve ?

There has been repeated feedback that the c4core submodule makes it impossible for some people to use rapidyaml, either because it contains submodules, or because c4core is a dependency boogeyman. 

While I disagree with both of these reasonings, it is not for me to dispute their choice.

So I am removing these reasons by making the c4core code really standalone in rapidyaml, while striving to retain all current existing workflows.

Also, narrowing down the amount of c4core code used by rapidyaml makes the standalone rapidyaml library marginally smaller, and enforces the standalone default of rapidyaml bypassing the submodule. 

---
See also #123


biojppm:
CI green! Merging now. There will be more follow-up PRs (eg swig interface is out of sync, and c4core version needs to be on master) but this is good enough for now.

PR #639, opened 2026-06-27
Fix clang warnings: -Wreserved-identifier (no logic changes)

biojppm:
Follow-up to #610


biojppm:
The failure in the coverage test is from touched lines that were not covered. Merging now.

Issue #656, opened 2026-08-11
resolve(): stack overflow on cyclic references, and unbounded alias expansion (DoS)


AlexandrKhromov2005:
Fuzzing `Tree::resolve()` under ASan (the existing `test/test_fuzz` harness exercises parse + emit but never calls `resolve()`) turns up two ways a tiny YAML input can take a process down when an application resolves untrusted aliases. Both are on current `master`; PoCs verified against the v0.16.0 single-header.

### 1. Stack overflow on a cyclic reference (8 bytes)

```cpp
ryml::Tree t = ryml::parse_in_arena("&x\n- *x\n");  // or "&x {a: *x}\n", or "&r [*r]\n"
t.resolve();                                         // SIGSEGV (stack overflow)
```

`resolve()` materializes each alias by duplicating its target subtree into the reference's location. When a reference's target is the reference node itself or an ancestor of it (a cyclic node), the target subtree *contains* the reference, so `Tree::duplicate()` / `Tree::duplicate_children()` recurse without bound:

```
#N Tree::duplicate            src/c4/yml/tree.cpp:965
#N Tree::duplicate_children   src/c4/yml/tree.cpp:988
#N Tree::duplicate            src/c4/yml/tree.cpp:965
... (unbounded) ...
```

The parser already accepts cyclic anchors — `TEST(anchors, circular)` parses exactly `&x\n- *x\n` — but nothing rejects them at resolve time. A stack overflow is a hard crash (uncatchable via the error callbacks). PR linked below adds cycle detection in the reference-matching pass and rejects it with a normal visit error, leaving valid resolution unaffected.

### 2. Unbounded alias expansion — "billion laughs" (sub-KB → OOM)

```yaml
l0: &l0 [x,x]
l1: &l1 [*l0,*l0]
l2: &l2 [*l1,*l1]
# ... up to l30
```

`resolve()` copies each alias target in full, with no bound on the total. Nested anchors expand multiplicatively: the 636-byte input above (a 94-node tree) expands to `2^30 ≈ 1.07e9` nodes; `resolve()` runs until it exhausts memory (`ERROR: [basic] could not allocate memory`, then abort under the default handler). Even `width=6, levels=6` (226 bytes) already produces 403,106 nodes.

[...]

PR #657, opened 2026-08-11
resolve: reject cyclic references instead of overflowing the stack

AlexandrKhromov2005:
Fixes the stack overflow in #656 (finding 1).

`Tree::resolve()` materializes each alias by duplicating its target subtree into the reference's location. When a reference's target is the reference node itself or one of its ancestors — a cyclic node such as `&x [*x]`, `&x {a: *x}`, or

```yaml
&x
- *x
```

— the target subtree contains the reference, so duplicating it recreates the reference inside the copy and `duplicate()`/`duplicate_children()` recurse without bound until the stack overflows. An 8-byte input is enough, and because it's a stack overflow it can't be caught through the error callbacks.

The parser already accepts these (see the existing `TEST(anchors, circular)`, which parses `&x\n- *x\n`); nothing rejected them at resolve time.

This adds a check in the reference-matching pass: once a reference's target is known, if the target is the reference itself or an ancestor of it, raise a visit error (`cyclic reference`) — the same mechanism used for other malformed reference situations. Non-cyclic references and valid anchor/alias resolution are unaffected.

A regression test is added next to `TEST(anchors, circular)`.

### Verification

- Built `src` + bundled `c4core` with the change and confirmed the three cyclic inputs above now raise the visit error instead of overflowing the stack, while valid anchored documents still resolve correctly (byte-identical output).
- Ran a ~210s libFuzzer + ASan campaign over `parse_in_arena` → `resolve()` → `emitrs_yaml` on the patched build, seeded with the crashing inputs the same harness found on unpatched `master`: the stack-overflow class is gone and no new crash appears.

Not addressed here: the unbounded alias-expansion (billion-laughs) DoS, also in #656 — that needs a separate expansion bound and is left for a follow-up.

### DIFF
--- a/src/c4/yml/reference_resolver.cpp
+++ b/src/c4/yml/reference_resolver.cpp
@@ -190,6 +190,11 @@ void ReferenceResolver::resolve_()
         if( ! refdata.type.is_ref())
             continue;
         refdata.target = lookup_(&refdata);
+        // A reference whose target is itself or one of its ancestors describes a cyclic node.
+        // Resolving it materializes the target subtree (which contains this reference) into a
+        // location inside that same subtree, recursing without bound. Reject it instead.
+        if(refdata.target == refdata.node || m_tree->is_ancestor(refdata.node, refdata.target))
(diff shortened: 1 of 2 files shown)


biojppm:
Thanks!

PR #662, opened 2026-09-05
Fix `c4_install_exports` call with system `c4core`

### LINKED ISSUE #659: Missing find_package() for c4core with RYML_SYSTEM_C4CORE=ON?
In a [downstream issue](https://bugzilla.redhat.com/show_bug.cgi?id=2517507), a user reported that when they attempt to build a project that uses `rapidyaml` via CMake `find_package` on Fedora 45 (currently in pre-beta development), they get:

```
CMake Error at /usr/lib64/cmake/ryml/rymlTargets.cmake:70 (set_target_properties):
  The link interface of target "ryml::ryml" contains:

    c4core::c4core

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

  /usr/lib64/cmake/ryml/rymlConfig.cmake:31 (include)
  cmake/SearchForStuff.cmake:24 (find_package)
  CMakeLists.txt:39 (include)
```

Fedora 45 has `rapidyaml` 0.16.0, with https://github.com/biojppm/rapidyaml/pull/653 as a patch.

The reporter notes that for `rapidyaml` 0.10.0 in Fedora 44, `/usr/lib64/cmake/ryml/rymlTargets.cmake` begins with:

```cmake
#-----------------------------
include(CMakeFindDependencyMacro)
find_dependency(c4core REQUIRED)
#-----------------------------
```

but there is no such `find_dependency()` call in the corresponding file in `rapidyaml` 0.16.0 in Fedora 45.

It seems like this might be an actual bug in the implementation of `RYML_SYSTEM_C4CORE` in `rapidyaml` 0.16. I don’t think that this is affected by anything I’m doing downstream, and I can reproduce the absence of a `find_package()` call in a git checkout:

```
$ git clone https://github.com/biojppm/rapidyaml.git
$ cd rapidyaml
$ cmake -S. -Bbuild -DRYML_SYSTEM_C4CORE:BOOL=ON -DBUILD_SHARED_LIBS:BOOL=ON
$ cmake --build build
$ DESTDIR="${PWD}/foo" cmake --install build
$ grep -FrinI find_package foo/usr/local/lib64/
foo/usr/local/lib64/cmake/ryml/rymlConfigVersion.cmake:1:# This is a basic version file for the Config-mode of find_package().
```

(Of course, the above requires a compatible version of `c4core` to exist as a system package. I used a Fedora 45 chroot.)

What do you think? Thanks in advance for your input.

musicinmybrain:
Fixes #659. Based on a suggestion by Sébastien Noel (@twolife) in a comment on that issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,7 +200,7 @@ endif()
 
 if(RYML_INSTALL)
     c4_install_target(ryml)
-    if(RYML_STANDALONE OR RYML_SYSTEM_C4CORE)
+    if(RYML_STANDALONE AND NOT RYML_SYSTEM_C4CORE)
         c4_install_exports()
     else()
         c4_install_exports(DEPENDENCIES c4core)


biojppm:
Thanks!

